PR #29, opened 2015-04-30
Attempt at a selenium test for submitting an URL

caprenter:
See what you think


Bjwebb:
Looks good!

Bjwebb:
I've also deployed this to http://dev.ocds.opendataservices.coop/ and the tests work successfully against that too.

Issue #25, opened 2015-04-30
Create some proper footer copy


caprenter:
The Footer needs some copy including link to github, etc

Bjwebb:
We might also want to link the version number to that version on GitHub, e.g. https://github.com/OpenDataServices/cove/tree/af4a19d

PR #37, opened 2015-05-06
[#25] Adds some more copy to footer. Links to git verison

caprenter:


Bjwebb:
I'm not 100% happy with some of the copy here and in the README, but I think it's better than having nothing, so I'll merge it in now and propose some changes later on.

Issue #59, opened 2015-06-18
Improve explore page for 360


caprenter:
When a file is uploaded and converted we could show a bit more useful info on that page

PR #60, opened 2015-06-18
[#59] Adds a table and unique id count to the 360 explore page

caprenter:
Just some basic stuff based on the basic stuff for ocds
Could do with working out how to handle dates as dates and not
simply as strings

### DIFF
--- a/cove/views.py
+++ b/cove/views.py
@@ -60,8 +60,45 @@ def get_releases_aggregates(json_data):
 def get_grants_aggregates(json_data):
     count = len(json_data['grants']) if 'grants' in json_data else 0
     
+    ids = []
+    unique_ids = []
+    #release_dates = []
+    #earliest_release_date = None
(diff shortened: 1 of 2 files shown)


cove/views.py
```diff
@@ -60,8 +60,45 @@ def get_releases_aggregates(json_data):
...
+            # Gather all the ocids
+            ids.append(grant['id']) if 'id' in grant else 0
+            
+            #Gather all the release dates
+            #release_dates.append(release['date']) if 'date' in release else 0
+            
+            # Some identifying data from a release that could be displayed in a table
+            generic_info = {}
+            if 'title' in grant:
+                generic_info['title'] = grant['title']
```
Bjwebb: Line 79 and 80 are maybe clearer written as:

```
generic_info['title'] = grant.get('title')
```

This is also true of the next two if statements (but not line 85).

Issue #83, opened 2015-08-05
360: Acceptable file formats - update guidance


stevieflow:
Cited:

> Acceptable files are:

>    Summary Spreadsheet - Excel
>    JSON built to the 360Giving Data Standard JSON schema
>    Multi-table data package - Excel

However, a CSV file seems to work well (I havent got other formats to work well just yet).  

Not sure what the solution is - but the current guidance seems inaccurate

Bjwebb:
Yes, single (but not multiple) CSV is an option for input (but not output), see https://github.com/OpenDataServices/cove/issues/62, so the guidance definitely needs updating for that.

Issue #82, opened 2015-08-05
Wrong link URL for 360 templates


stevieflow:
On http://cove.opendataservices.coop/360/

> The application accepts data in some of the formats given in the 360Giving Data Standard publishing guidence under Choose your template.  

these links are not direct any longer

caprenter:
Thanks - yes I agree. I'll get it fixed.

Issue #16, opened 2015-04-29
Test should not call external urls


Bjwebb:
Where possible we should be using URLs created locally, so that running the tests doesn't depend on having an internet connection, and the availability of third party services.

Bjwebb:
https://pypi.python.org/pypi/pytest-localserver may be a useful library to help with this.

Bjwebb:
The remaining of external test urls are those of schemas. These are also fetched externally in the application proper, so I have opened a new issue about this https://github.com/OpenDataServices/cove/issues/52

Bjwebb:
https://github.com/OpenDataServices/cove/pull/85 possibly introduces more dependency on external urls to FTS tests. However, I'm not sure if the FTS tests have ever passed without an internet connection,  due to use of external bootstrap.

Bjwebb:
http://django-offlinecdn.readthedocs.org/en/latest/ is a possible approach to continuing to use an external CDN for bootstrap, but also having an offline copy for development.

Bjwebb:
https://github.com/kevin1024/vcrpy is also an interesting approach of capturing what requests to external urls look like, and then automatically mocking them for future test runs.

robredpath:
Closing out - it's pretty rare that we need to run tests entirely offline. We're also refactoring our tests to make it easier to run subsets of tests as necessary.

PR #85, opened 2015-08-10
[#82][#83] Fixes up links from cove to the new 360 site

caprenter:
Also alters the main site page title to CoVE
Adds tests to check links go to appropriate URL
Checks footer link to see that they actually click through to the
correct sites and see appropriate text. These might not be all that clever
in the longer term, but should flag significant changes perhaps.
Re-runs translastions as some text has been altered

### DIFF
--- a/cove/settings.py
+++ b/cove/settings.py
@@ -44,7 +44,7 @@ COVE_CONFIG_BY_NAMESPACE = {
     'application_name': {
         'cove-ocds': _('Open Contracting Data Tool'),
         'cove-360': _('360Giving Data Tool'),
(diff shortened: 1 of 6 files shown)


caprenter:
@Bjwebb just realised that some of the new tests conflict with this ticket: Test should not call external urls
- https://github.com/OpenDataServices/cove/issues/16

Bjwebb:
Merging this as in general it's an improvement. Will add a note to the ticket about fts tests (which I'm not sure have ever worked without external urls, due to use of bootstrap).

Issue #81, opened 2015-08-04
Functional tests fail against the dev site


Bjwebb:
(... but work against the live site)

ie. `CUSTOM_SERVER_URL=http://dev.cove.opendataservices.coop/ py.test` fails but `CUSTOM_SERVER_URL=http://cove.opendataservices.coop/ py.test fts` passes.

This is because the dev site inserts a different banner into the page, which breaks the tests that look for it.

caprenter:
We discussed this, and the failure is not because of the banner.

In fact both banners are present on the dev site, but one sits above the other with a css z-index thingy.

The failing tests are to do with trying to check that there is an appropriate message for someone coming back to view a file that has been deleted. Because the databases are different on live and dev, this is hard to do. 

The banner check is kind of appropriate. It can test the dev site, but as that banner is added via the server and not this code, maybe we don't need it.

Bjwebb:
> The failing tests are to do with trying to check that there is an appropriate message for someone coming back to view a file that has been deleted. Because the databases are different on live and dev, this is hard to do.

That's not quite true.

In fact the failing tests are checking that data that's never existed gives the correct error. It should be quite straightforward to change to an identifier that's never existed on the live or dev sites.

PR #86, opened 2015-08-10
[#81] Should fix tests against appropriate banner for  live or dev

caprenter:
By checking the server_url in the banner test we should get the
correct behaviour?


Bjwebb:
This is not the reason that the tests were failing. I believe this is still open because we discussed additionally testing for the dev banner on dev.

Bjwebb:
I've rebased and editted this just to do the additional test if it's the dev site.

Issue #43, opened 2015-05-07
A failed conversion should show a custom error page


Bjwebb:
Currently this is Django's default error page.

The traceback should still be accessible for debugging purposes. For now, it is probably easiest to emit this to stdout.

Additionally, it would be useful to display extra information for a JSON file, even if the conversion to spreadsheet failed, so maybe the error should be part of the explore.html template.

Bjwebb:
We now have custom errors for badly formed JSON https://github.com/OpenDataServices/cove/issues/51, but there are other ways the conversion could fail.

Bjwebb:
I've done some basic work on this for conversion from a spreadsheets, but it still needs tests.

Also, it needs to be added for JSON conversion (preferably still showing the rest of the page), and we need to still record the errors somehow. Using sentry's integration with the  `logging` module looks like it might be useful for this - http://raven.readthedocs.org/en/latest/integrations/django.html#integration-with-logging

Bjwebb:
There is now a similar custom page for JSON conversion, and tests for both.

The page for JSON conversion is not a normal error page, as it shows the other information about the JSON file - see http://43-catch-conversion-errors.dev.cove.opendataservices.coop/ocds/data/9809fe80-3658-43e3-a5b6-9f1eddec3459 for example. Currently it indicates that something is wrong by turning the Download Files box yellow instead of green. Considering currently the other errors are blue, there may be some inconsistency here (which I'll likely flag in a new issue).

Remaining task is to determine how best to log these errors, now that they won't automatically go to sentry.

Bjwebb:
In https://github.com/OpenDataServices/cove/commit/6daf89c1c944ac9b996ae1dfec2760a02a8192a9 the captured errors are now logged to sentry.

I've also configured sentry to not send notifications for these errors (matching on what files doing the logging), to reserve notifications for completely handled exceptions.

PR #91, opened 2015-08-25
[#43] Catch conversion errors

Bjwebb:
When a conversion fails, we now get a custom error page, that 1) is themed nicely, 2) displays the conversion error to the user, 3) reports the conversion failure to sentry (but doesn't notify). In the case of a JSON file supplied, the rest of the page is also shown.

See https://github.com/OpenDataServices/cove/issues/43 for more info.

Examples:
Spreadsheet conversion: http://43-catch-conversion-errors.dev.cove.opendataservices.coop/360/data/d321c57c-6b7f-4945-8e8d-4343f28678f5
JSON conversion: http://43-catch-conversion-errors.dev.cove.opendataservices.coop/ocds/data/9809fe80-3658-43e3-a5b6-9f1eddec3459


Bjwebb:
@caprenter May be good to discuss this briefly.

Issue #127, opened 2015-10-06
In functional tests, check that linked files actually exist


Bjwebb:

Bjwebb:
This should include converted files, and the link to the original.

PR #136, opened 2015-11-05
[127] Check files actually exist.

kindly:
New checks to see firstly if the href attribute has the correct file names in. Secondly to actually request the files to see if they return a 200 and have some size.


Bjwebb:
`[127]` doesn't have a `#` in it, so isn't being picked up as an issue number. Should probably be `[#127]`. Could you amend the commit?

Issue #142, opened 2015-11-16
Update outdated requirements


Bjwebb:
https://requires.io/github/OpenDataServices/cove/requirements/?branch=master

The versions in https://github.com/OpenDataServices/cove/blob/master/requirements.txt and https://github.com/OpenDataServices/cove/blob/master/requirements_dev.txt need updating, and then we need to check that tests run etc.

idlemoor:
Thanks Ben, I'll block out Wednesday (18th) for this if that's ok?

Bjwebb:
Sounds good.

idlemoor:
For the record, the updates were
(live) Django==1.8.6 requests==2.8.1 raven==5.8.1 jdcal==1.2 pytz==2015.7 schema==0.4.0 sqlparse==0.1.18
(dev) flake8==2.5.0 coverage==4.0.2 Werkzeug==0.11.2 coveralls==1.1

PR #143, opened 2015-11-19
[#142] Update Python requirements

idlemoor:
Completely straightforward upversion, just like requires.io told us. This was completely painless -- there's got to be some snag?


Bjwebb:
Looks good. Sometimes life really is that simple (famous last words).

Issue #46, opened 2015-05-08
Show error messages when JSON doesn't validate against the schema


Bjwebb:
This builds upon https://github.com/OpenDataServices/cove/issues/9

This may want its own page, linked to from the validation box on the main explore page.

Bjwebb:
@kindly Not sure how clear the title is, but this is the issue for having more than just "Yes" or "No" as validation information in Cove.

PR #146, opened 2015-11-19
[#46] Add validation error table.

kindly:
Add table within a panel that shows frequency and location of first 3 validation errors for each error message.

Refactor functional tests slighlty so that you can provide a list of strings to be found in page instead of needing for the whole test to be run again for different string.


Bjwebb:
Not had chance to review this, but I think other members of the team would like to look at/use this in the meantime, so I've deployed it to http://46-validation-errors.dev.cove.opendataservices.coop/

Issue #147, opened 2015-11-19
Allow OCDS records


kindly:
Need to allow record packages, not just release packages. 

This will require an alteration to the explore page and a way to supply multiple validators.

PR #154, opened 2015-11-25
[#147] ocds allow records

kindly:


Bjwebb:
Tests fail when I try merging this because records don't play nicely with the new validation code: https://travis-ci.org/OpenDataServices/cove/builds/94203913

@kindly Could you look at rebasing and fixing this?

Issue #155, opened 2015-11-25
Validation Speedup and duplicate reporting


kindly:
Validating is slow at the moment with large files.

This is mainly due (after doing some profiling) to the jsonschema validator "uniqueItems".   When using this validator it compares the whole object, which is not hashable and therefore it has to do a brute force O(n^2) compare.  So for large files this will get slower and slower.

The suggestion is that for our case we just need to compare the "id" field in these objects.  This also means we can easily give reports of ids that are duplicates, not just whole objects.

Also patternPropperties is slow and mostly unnecessary for our use case.

PR #156, opened 2015-11-25
[#155] Validation speedups.

kindly:
Replace uniqueItems validator with one that checks for unique ids.
Remove patternProperties validator.

### DIFF
--- a/cove/views.py
+++ b/cove/views.py
@@ -11,12 +11,33 @@ import collections
 from flattentool.json_input import BadlyFormedJSONError
 import requests
 from jsonschema.validators import Draft4Validator as validator
+from jsonschema.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


Bjwebb:
Not everything with `uniqueItems` has `id`s e.g. `Organization`:
https://github.com/open-contracting/standard/blob/master/standard/schema/release-schema.json#L310
https://github.com/open-contracting/standard/blob/master/standard/schema/release-schema.json#L688

This change would mean no uniqueness check for Organization.

kindly:
I was aware of this. Thought the benefits and simplicity would outweigh the slight loss in validation completeness. 

However, I have made a fix to fall back to original validation if there are any item in the array without an id, which means you will also get uniquness validation errors if there are missing ids.

Issue #92, opened 2015-08-26
Load new data button


stevieflow:
After submitting some data (360 version) I cant locate a "load new data" button (Im used to seeing that in the IATI validator)

PR #177, opened 2015-12-09
[#92] Adds a load new data button

caprenter:
Should display on all explore pages, and test has been written for this
Slightly repositioned the language drop down to align with button
Regenerated .po files as text in button can be translated

### DIFF
--- a/cove/static/dataexplore/css/style.css
+++ b/cove/static/dataexplore/css/style.css
@@ -8,7 +8,7 @@ h1.application-name {
 }
 
 .language-select {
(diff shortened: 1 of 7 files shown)


Bjwebb:
@caprenter This is now out of sync with master, could you have a go at rebasing, and assign to @kindly when you're done / get stuck.

kindly:
I am looking into this as the merge is quite tricky and I have a better sense of it.

caprenter:
This is great - does it need assigning to @Bjwebb to merge?

PR #178, opened 2015-12-09
Adds a tablesorter to the grants table

caprenter:
I just kind of think we need this - even if the table doesn't stay forever


caprenter:
I think @kindly is looking at implementing this in a similar/different way, so I'll assign to him, and he can close this pull request once he's done his own thing

caprenter:
There is an alternative approach in https://github.com/OpenDataServices/cove/pull/190 so this can now be closed

PR #184, opened 2015-12-10
168 which schema

caprenter:
Presented for discussion, not necessarily for merging

### DIFF
--- a/cove/templates/explore.html
+++ b/cove/templates/explore.html
@@ -75,7 +75,7 @@
     <div class="panel panel-danger">
       <div class="panel-heading">
         <h4 class="panel-title">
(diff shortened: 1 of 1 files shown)


caprenter:
NB: The text in the panel footer is not in translation blocks, and no tests have been written to detect the presence of these messages.
The change does apply to both 360 and OCDS explore views
It uses the URL of the schema. This is probably not good - but I'm not actually sure that we can do anything else. (e.g. the ID in the ocds schema is a url, 360 doesn't have anything really)

kindly:
This should be done as part of #242

Bjwebb:
Done https://github.com/OpenDataServices/cove/pull/309

Issue #189, opened 2015-12-11
Add Data Tables for Grants


kindly:
Add a nice way of searching and sorting grants

PR #190, opened 2015-12-11
[#189] Add data tables for grants.

kindly:
The data tables plugin works out of the box and gracefully degrades with no js.


Bjwebb:
When I try this with http://www.wolfson.org.uk/media/2740/wolfson-foundation-grants-2014.xlsx the table is blank until I click on page.

Bjwebb:
Additionally, it would be good to add a basic tests that this loads correctly, and that the text we expect (ie one of the first 10 rows) is still shown.

kindly:
The issue with that file is that it has ten empty grants in it.  Flattentool seems to be converting empty rows in a spreadsheed to empty grants '{}'. 
But, agreed there should be a test or two.

Bjwebb:
Thanks for explaining. There's an issue for this in flattentool at https://github.com/OpenDataServices/flatten-tool/issues/80

PR #192, opened 2015-12-11
[#192] No flattentool and validation on reload

kindly:
Check if flattened/unflattened file exists and if they do no rerun flattentool.
Save validation results to disk.


Bjwebb:
This looks good.

Would be nice to see if we could modify some of our tests to check the page a second time, to make sure everything works as expected.

Issue #193, opened 2015-12-14
unorderable types: int() > str()


Bjwebb:
Error in `get_grants_aggregates` when looking at this file:
http://www.wolfson.org.uk/media/2738/wolfson-foundation-grants-2014.csv

Bjwebb:
It would be good to have some tests to prevent this sort of thing. The most basic test would be to test for the particular case that was the problem here. However, I wonder if there's some broader testing we could do with a something like [hypothesis](https://hypothesis.readthedocs.org/en/latest/).

At any rate, I think how we test this could benefit from some thought, and probably some back and forth, so I've merged the existing pull request, so we don't feel under pressure to 'get the tests done' before that goes in.

Bjwebb:
I think @kindly has been looking into using a random JSON schema instance generator for this, which is a good approach.

Part of the reason I mention hypothesis, is that I had a brief look at setting it up for flatten-tool during PyConUK - https://github.com/OpenDataServices/flatten-tool/compare/hypothesis

Bjwebb:
We now have the random JSON schema instance generator tests https://github.com/OpenDataServices/cove/commit/db53be266e671e7054eb968702fa0483fcd5296a

PR #197, opened 2015-12-15
[#193] Check types for grant aggregates.

kindly:
Make sure we only use values of the correct type.


Bjwebb:
Merging this to get it fixed, but see my comment about tests at https://github.com/OpenDataServices/cove/issues/193#issuecomment-165090246

PR #227, opened 2016-01-07
[#226] Config to allow change of prefix.

kindly:
For example:
PREFIX_MAP=ocds=validator/,360=360/
will change ocds to be at validator and keep 360 the same.
PREFIX_MAP=ocds=validator/
will remove 360 routes entirely.


Bjwebb:
Looks good. https://github.com/OpenDataServices/cove/issues/228

I think there's some testing to be considered here, but I've created a new issue for that https://github.com/OpenDataServices/cove/issues/228

Issue #230, opened 2016-01-08
Add button to flatten json.


kindly:
Do not flatten automatically.

Bjwebb:
This is now implemented on the dev site: http://dev.cove.opendataservices.coop/

PR #231, opened 2016-01-08
[#230] Button to flatten

kindly:
Mostly test fixes to allow for this.


Bjwebb:
Some conflicts to resolve now.

Also, it looks like there's a bullet point ontop of the K in KB. I'm guessing this should be hidden/avoided instead?
misplaced_bulled

kindly:
The conflicts has been resolved.
Firefox has no issues with the above and I have spent ages trying to get both working, but gave up.  I have now got rid of the bullet points all together, which I prefer anyway.

Issue #228, opened 2016-01-08
Testing alternate endpoints


Bjwebb:
Currently it's possible to run ` CUSTOM_SERVER_URL=http://cove.opendataservices.coop/ py.test fts` to test the deployed site. Following https://github.com/OpenDataServices/cove/pull/227, it would be good to be able to have a way of testing a site with a non-default prefix map (e.g. the deployed ocds validator).

PR #232, opened 2016-01-18
[#228] Test alternative endpoints

kindly:
Allow you to add enironment variables PREFIX_360 and PREFIX_OCDS to test
endpoints that are not in the normal locations.

There is a also a slight change to config so that if you do not put in
any SENTRY config then it does not use it.


Bjwebb:
Some conflicts on this now. Looks good though, we can get this merged when we meet tomorrow.

Issue #236, opened 2016-01-21
Add transifex code and document usage.


kindly:

kindly:
fixed in #239

kindly:
fixed in #239

Issue #237, opened 2016-01-21
Limit the translations in Cove to only Spanish


Bjwebb:
(All our translations are machine translated and incomplete. Spanish is the only one we have realistic plans to look after in the immediate future).

kindly:
This is done in #236

PR #239, opened 2016-01-21
[#236] Add transifex support

kindly:
* Adds a .tx file for transifex cli
* Adds a en directory as a source
* Updates the readme.
* Removes the ro and fr translations
* Updates the tranlation files as they stand


Bjwebb:
You might want to reference https://github.com/OpenDataServices/cove/issues/237 about removing the translations, as well as removing them from settings.py as we just discussed.

Issue #238, opened 2016-01-21
Redo Spanish machine translation


Bjwebb:
I'll feed this through google translate again.

Should give me chance to try uploading to transifex too.

Bjwebb:
We discussed how transifex can use the google translate API to do this.

Bjwebb:
Now on dev http://dev.cove.opendataservices.coop/ocds/

PR #241, opened 2016-01-29
[#238] Spanish machine translate and fixes

kindly:
Fix html to work with google translate and add the spanish translation
from transifix.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -115,7 +115,7 @@ In order to generate messages and post them on transifex:
 In order to fetch messages from transifex:
 
 .. code:: bash
-    tx pull -t -a
+    tx pull -a
 
 In order to compile them:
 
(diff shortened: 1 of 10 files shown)


Bjwebb:
https://github.com/OpenDataServices/cove/pull/241/files#diff-4b7e6fde07397056e485f0601e1327dfR12

Moving variables outside of the blocktrans may cause trouble for some languages: e.g.
```diff
-    <h2 class="heading-in-panel" ><small>{% blocktrans %}How to use the {{ application_name }}{% endblocktrans %}</small></h2>
+    <h2 class="heading-in-panel" ><small>{% blocktrans %}How to use the {% endblocktrans %}{{ application_name }}</small></h2>
```

In some languages the word order will be different, e.g. so that the applications name must be at the start or middle of that sentence.

I don't think this is a problem for Spanish though, so we may want to decide to make this change for now. Has the potential to cause us trouble later though.

Issue #250, opened 2016-02-02
Commit link in footer is broken


Bjwebb:
http://dev.cove.opendataservices.coop/ocds/

"Running version "

PR #257, opened 2016-02-08
[#250] Add missing revision

kindly:
using revision.tag seems to work. It will use revision number if there is not a tag.   I have no idea where version came from.


Bjwebb:
@kindly This looks good btw, do you want to rebase and merge?

Issue #84, opened 2015-08-06
Feature Request: CoVE instances - supply sample file?


stevieflow:
For 360 and OCDS it would be useful to include a sample data file for people to "click thru"...

Bjwebb:
Yes. Might also be useful to link to a larger set of data.

For OCDS we have sample-data - https://github.com/open-contracting/sample-data

For 360Giving we have http://data.threesixtygiving.org/ , but not sure what the best specific sample file is. @stevieflow Do you have any thoughts?

caprenter:
We do have sample data used in our tests:
https://github.com/OpenDataServices/cove/tree/master/cove/fixtures

Bjwebb:
I think the considerations here might be a bit different than our unit tests. eg. It might be better to have a link to data that's kept up to date, and has more than two records.

I'm also not sure what the considerations are about pointing to a particular publishers data in the user interface.

Bjwebb:
I suggest that for OCDS we link to http://standard.open-contracting.org/validator/?source_url=https://raw.githubusercontent.com/open-contracting/sample-data/master/fictional-example/ocds-213czf-000-00001-02-tender.json for now (although I suspect we can improve on this).

For 360 I don't think we have appropriate issue yet, so I've opened a separate issue for that https://github.com/ThreeSixtyGiving/standard/issues/119

Bjwebb:
Pull request for the OCDS link https://github.com/OpenDataServices/cove/pull/271

robredpath:
@ekoner is working on some 360 sample data in https://github.com/ThreeSixtyGiving/standard/issues/119, so once that's ready we can add a link to the 360 validator

stevieflow:
I added some sample 360 files I used on a recent demo: https://github.com/ThreeSixtyGiving/standard/issues/119#issuecomment-265805900

Bjwebb:
@stevieflow Is that sample data suitable for linking to from 360 Cove? Should it be hosted somewhere more authoritative than attached to a GitHub issue?

stevieflow:
@Bjwebb yes please - I just added it to the ticket for reference.  Agree it should be somewhere more authoritative

caprenter:
Candidate for #531

Bjwebb:
@caprenter BTW OCDS has a "try loading some sample data." link now, the remaining task on this issue is for 360Giving.

Bjwebb:
[...]

PR #271, opened 2016-02-15
[#84] Add link to CoVE using OCDS sample data

Bjwebb:


Bjwebb:
@kindly Could you check you're okay with this?

PR #301, opened 2016-03-15
[#296] Catch stats producing errors.

kindly:
Catch all main errors caused when decending json objects and do not
show stats if this is the case.  Give warning to user saying this.

Errors will still be raised if the data passes validation but
there is an error in the stats code. So basically we do a good attempt
at producing stats for invalid data but only if the data is close
to being valid.


Bjwebb:
Could you rebase this and fix the flake8 errors? Then I think this'll be ready to merge.

Issue #44, opened 2015-05-07
Show warnings encountered during conversion


Bjwebb:
This may want its own page, much like https://github.com/OpenDataServices/cove/issues/46

Bjwebb:
Next steps:
* Investigate broken tests: https://travis-ci.org/OpenDataServices/cove/builds/119597153
* Write tests for the list of warnings

Bjwebb:
Existing tests are now fixed.

Still need to test displaying the warnings. This is tricky because it depends on the behaviour of flatten-tool (which we've deliberately separated out for a reason). One approach might be to mock the relevant flatten-tool warnings to return predictable warnings that we can test for.

Bjwebb:
Pull request https://github.com/OpenDataServices/cove/pull/321

PR #321, opened 2016-04-04
[#44] Capture conversion warnings and show them on the explore page

Bjwebb:
Creating a pull request to make it easier to review this.

This definitely wants some tests, but it may be worth merging into master before those are written, to give us longer to demo/test it manually.


Bjwebb:
Tests written now.

Issue #149, opened 2015-11-23
Repository is missing a CONTRIBUTING document


Bjwebb:
https://github.com/OpenDataServices/developer-docs/blob/master/contribute.md#contributing
https://github.com/OpenDataServices/developer-docs/blob/master/CONTRIBUTING.md

This came up when talking with @kindly in London.

PR #342, opened 2016-04-27
[#149] Adds a contributing document

caprenter:
This is based on a re-write of our generic information that has
not yet been agreed


Bjwebb:
@caprenter Thanks! This looks good to me.

I'm happy to merge, but do we want to run it by anyone else first?

Issue #207, opened 2015-12-16
Calling Cove as a Python library (without the web interface)


Bjwebb:
Some data publishers want to be able to use Cove's additional/custom checks in their tests.

We also have a usecase for making various checks of the sample data https://github.com/open-contracting/sample-data/issues/11 https://github.com/open-contracting/sample-data/issues/7

kindly:
The particular library functions we should consider are.

- Validation Errors  (in cove we modify a few of the json schema validation to be more suitable to the schema)
- Additional Fields (the report as to extra fields the data has)
- Flattentool warnings (possible errors in the conversion between the spreadsheets and the json)
- Summary Statistics  (aggregations of summary statistics of the data)
- Unflattening/Flattening (this could be done in flattentool directly but it would be really useful to have the config options set up for you in the exact way cove does)
- Any new feature that is appropriate should have a relatively easy path to becoming a library function.

The one thing that is uncertain is the how much we should reuse the main config file for this library and how we allow changing options within it in the library (i.e what schema we choose) should be managed.

Bjwebb:
As I understand it @robredpath has done some good refactoring to help with this, but we still need to think about some of the considerations above for providing a useful API. One way to proceed on this is to use the API ourselves outside of Cove - one example would be https://github.com/open-contracting/sample-data/issues/7

http://cove.readthedocs.io/en/307-sphinx-docs/lib/ is possible an overview of what functions exist in cove.lib now.

timgdavies:
Building on Ben's comment above, we have a use-case coming up again for OCDS sample-data repository, where @annapowellsmith is currently working on collating large corpus of published OCDS data for analysis. 

In this process it would be very useful to have the ability to use the full features of the OCDS validator at the command line to:

* Detect release or record package, version and extensions;
* Validate against the appropriate schema;
* Report on extra fields, codelist validation etc.

I can see that release-level validation could end up being very desirable here too (i.e. reporting on which releases fail validation), but I realise that is a separate issue.

robredpath:
@timgdavies I think that 'CoVE as a library' is a somewhat different proposition from 'CoVE on the cmdline' - there may be plenty of overlap in how we implement them, however. 

How much flexibility do we have over the interface? Does this need to be a neat cmdline tool from the off, or is it OK to use Django's manage.py and require the user to be running in a virtualenv to use it?

timgdavies:
[...]

PR #355, opened 2016-05-17
[#207] Refactor views.py and start to remove Django deps 

robredpath:
Satisfy #207

### DIFF
--- a/cove/views.py
+++ b/cove/views.py
@@ -9,9 +9,8 @@ from jsonschema.validators import Draft4Validator as validator
 from django.db.models.aggregates import Count
 from django.utils import timezone
 from datetime import timedelta
-
-from cove.lib.common import *
-from cove.lib.tools import *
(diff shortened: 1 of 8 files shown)


cove/views.py
```diff
@@ -2,609 +2,30 @@
...
-    format_checker = FormatChecker()
-    if current_app == 'cove-360':
-        format_checker.checkers['date-time'] = (datetime_or_date, ValueError)
-    for n, e in enumerate(validator(schema, format_checker=format_checker).iter_errors(json_data)):
-        validation_errors[e.message].append("/".join(str(item) for item in e.path))
-    return dict(validation_errors)
-    
-
-class CoveInputDataError(Exception):
+class CoveWebInputDataError(CoveInputDataError):
```
robredpath: I don't like this solution, but I'm not sure how else to achieve the same end result with a more generic exception

PR #376, opened 2016-05-25
Adds a common errors page to the site

caprenter:
First stab at this


Bjwebb:
This also creates an OCDS common errors page, which says 360Giving all over it http://localhost:8000/ocds/common_errors

We need to fix this before this eventually goes live, even if it's just by disabling the links for the OCDS validator.

kindly:
closed as this is done in #384

Issue #349, opened 2016-05-16
When I upload a file to CoVE, I want some feedback on my Recipient Org identifiers


stevieflow:
(this may or may not be unique to 360)

When I submit data to CoVE, I'd like my organisation identifiers to be "checked".  By this, I mean:

- ``Recipient Org: Identifier`` is this populated?  If blank - please tell me (which I think is part of schema validation anyway)

- ``Recipient Org: Identifier`` is this prefix present on the IATI Org Agency codelist: http://iatistandard.org/202/codelists/OrganisationRegistrationAgency/ ? 

- ``Recipient Org: Identifier`` - is the org number "valid" ? *

* Realistically, we probably want to only focus on England, Wales, Scotland and NI companies and charities, for 360Giving

NB: In 360Giving there are two additional columns for company and charity numbers.  We could dedicate resource to checking these numbers too, but I don't think so.  This can be especially complex as there is no means to declare if this number is on the England & Wales or Scotland register of charities, for example - whereby at least via ``Recipient Org: Identifier`` there are documented steps for publishers to take

robredpath:
- regex check the field
- no external lookups
- feed back on if some grants are missing the recipient org block at all 
- check prefix against IATI org codelist

robredpath:
- report on unique prefixes, how many of each type, etc

Bjwebb:
Assinging me first of all, as I have knowledge about the IATI codelists, and I think this might relate to https://github.com/OpenDataServices/cove/issues/361 too.

stevieflow:
@Bjwebb useful issue/content from @ekoner  at https://github.com/OpenDataServices/grantnav/issues/19

Bjwebb:
@stevieflow Thanks for the link. Just to clarify, we're not thinking of performing lookups against external registers in this sprint, so I don't think the content of that issue is relevant yet.

Bjwebb:
I'm planning on implementing this in `get_grants_aggregates` in https://github.com/OpenDataServices/cove/blob/master/cove/lib/threesixtygiving.py and then make some additions to the template along the lines of the current "Key field information" box.

Bjwebb:
> - ``Recipient Org: Identifier`` is this populated?  If blank - please tell me (which I think is part of schema validation anyway)

Yes, this is part of schema validation. There was a bug where this didn't fail if there was no recipient org information at all, but this is now fixed https://github.com/OpenDataServices/cove/issues/360

[...]

PR #386, opened 2016-05-27
[#349] Display recipient org id prefixes in key field info

Bjwebb:
Also displays any recipient org IDs that don't have a recognised prefix.

No tests for this yet, but we may want to merge already to get some feedback on it along with the rest of our changes this week.


kindly:
We need to add the 360- prefix to the allowed list.

Bjwebb:
@kindly btw, it's `360G-` (with the G), see http://www.threesixtygiving.org/standard/identifiers/#toc-creating-identifiers

Issue #347, opened 2016-05-09
As a developer, I should be able to run CoVE's tests in parallel where appropriate


robredpath:
Running cove's tests on my laptop (in my VM!) takes ~12mins. TravisCI takes 8 mins to run them. If we can get that down, it becomes quicker and easier to test multiple iterations.

robredpath:
@kindly @Bjwebb @caprenter Any objection to using pytest-xdist (https://pytest.org/latest/xdist.html) ? Tried it out on my dual-core VM and it halved execution time. All the tests passed. I had a read of the tests and couldn't see any that would have issues running in parallel, but you guys know them better than me.

Bjwebb:
pytest-xdist looks like a sensible choice.

The tests *should* each be self contained such that running in parallel would be fine.

I'd be happy to see this enabled.

robredpath:
I've given it a try locally and it's great - halves execution time with just -n2. How do I add this - there's two requirements.txts, and the .in files?

Bjwebb:
Yes, how we do requirements needs documenting in the README. I'll look into doing this today.

Bjwebb:
I've updated the script, and added some info to the README in https://github.com/OpenDataServices/cove/commit/f87c8f13eb84cb3a81086463235b43bda7d0dfe1

Hopefully that makes it a bit clearer how to do what you want. In this case `requirements_dev.in`is the one to add for, as we don't run the tests on the live deployed copy.

PR #407, opened 2016-06-02
[#347] Add xdist to dev reqs for parallel testing

robredpath:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
 
 script:
   - "flake8 ."
-  - "py.test --cov cove"
+  - "py.test --cov cove -n 2"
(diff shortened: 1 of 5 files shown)


caprenter:
This will probably/may pass if you grab the fix in:
https://github.com/OpenDataServices/cove/commit/f331476c6f3dfb016822b0b471208ff78cbd3e38

Bjwebb:
Looks like this has been rebased, and then merged with the non-rebased copy. Might be best to undo that merge?

robredpath:
@Bjwebb I've squashed anyway, which I think solves the problem

Bjwebb:
@kindly I've made some changes here to ensure that the requirements for this are in the pinned file. I think it would be good for you to review and merge, or for us to discuss tomorrow morning.

Bjwebb:
This is failing intermittently on Travis and on my local machine. @robredpath Could you take a look at this?

Bjwebb:
I think I've realised what's wrong here, no all our tests load the page they're intending to test, instead relying on the page that was loaded by the previous test. This causes inconsistent failures with xdist, because which set of tests a particular browser will be scheduled will vary. Fix is to make sure our tests load the pages they need. I'll take a look at that.

Bjwebb:
Looks like it works. \o/

Issue #404, opened 2016-06-02
Funding Org identifiers prefix not checked


caprenter:
I saw this the other day and have been working on a suggestion for it. Need an issue to place it against

PR #410, opened 2016-06-02
[#404] Reports on funding org prefixes

caprenter:
This started off as trying to see why funding org prefixes were not
picked up as failing, and ended up being a copy and paste job of the
receiver org prefixes work.

See if it is useful.


caprenter:
This would also fix: https://github.com/OpenDataServices/cove/issues/412

PR #415, opened 2016-06-06
[#413] All fts not rerun for checking refresh and param file input

kindly:


kindly:
This makes the test almost run twice as fast,

Bjwebb:
Looks good. Might be good to have an example of each type of file (JSON,CSV,XLSX) in that list?

kindly:
@Bjwebb I have changed them as you suggested.

Issue #472, opened 2016-07-26
conversion error should be conversion warning


caprenter:
When we get a conversion warning, we are displaying this as 'conversion error'

PR #473, opened 2016-07-26
[#472] Alters conversion errors text to conversion warnings

caprenter:
Changes the text
Updates the tests

### DIFF
--- a/fts/tests.py
+++ b/fts/tests.py
@@ -449,9 +449,11 @@ def test_flattentool_warnings(server_url, browser, httpserver, monkeypatch, warn
     body_text = browser.find_element_by_tag_name('body').text
     if len(warning_texts) == 0:
         assert 'Conversion Warnings' not in body_text
+        assert 'conversion errors' not in body_text
     else:
         assert warning_texts[0] in body_text
         assert 'Conversion Warnings' in body_text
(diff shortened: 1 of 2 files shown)


fts/tests.py
```diff
@@ -449,9 +449,11 @@ def mockflatten(input_name, output_name, *args, **kwargs):
     body_text = browser.find_element_by_tag_name('body').text
     if len(warning_texts) == 0:
         assert 'Conversion Warnings' not in body_text
+        assert 'conversion errors' not in body_text
     else:
         assert warning_texts[0] in body_text
         assert 'Conversion Warnings' in body_text
+        assert 'conversion errors' not in body_text
```
Bjwebb: These are useful, but it might be good to have both of them be case insensitive, e.g.:
```
         assert 'Conversion Warnings' in body_text.lower()
         assert 'conversion errors' not in body_text.lower()
```

caprenter: Conversion Warnings is the exact text used as the heading in the Conversion Warnings panel (L110). 
conversion errors is the exact text we have replaced.

A better test to see that 'conversion warnings' and 'Conversion Warnings' are both shown could be used, but I figured that what we have now would be good enough.

Issue #482, opened 2016-09-29
CoVE does not recognise files with uppercase extensions


ekoner:
Uploading a file with the extension XLSX fails with this message:

> Sorry we can't process that data
> 
> We did not recognise the file type.
> 
> We can only process json, csv and xlsx files.

See http://dev.cove.opendataservices.coop/360/data/059bcf0f-9f12-43e1-b19f-fe6f3ca14116

File added to the **Bad Data** folder on Google Drive

Bjwebb:
This is fixed in master now, and on the dev site http://dev.cove.opendataservices.coop/

PR #483, opened 2016-10-03
[#482] Recognise uppercase file extensions

Bjwebb:


Bjwebb:
(build fails because use-mirrors has been removed)

Issue #476, opened 2016-08-04
Badly typed IDs cause server error


Bjwebb:
### Non-unique IDs that are integers

https://app.getsentry.com/open-data-services/cove/issues/144605197/

```
TypeError: sequence item 0: expected str instance, int found
(4 additional frame(s) were not displayed)
...
  File "cove/views.py", line 35, in wrapper
    return func(request, *args, **kwargs)
  File "cove/views.py", line 151, in explore
    validation_errors = common.get_schema_validation_errors(json_data, schema_url, schema_name, request.current_app, cell_source_map, heading_source_map) if schema_url else None
  File "cove/lib/common.py", line 182, in get_schema_validation_errors
    for n, e in enumerate(validator(schema, format_checker=format_checker, resolver=CustomRefResolver('', schema, schema_url=schema_url)).iter_errors(json_data)):
  File "cove/lib/common.py", line 41, in uniqueIds
    yield ValidationError("Non-unique ID Values (first 3 shown):  {}".format(", ".join(list(non_unique_ids)[:3])))

TypeError: sequence item 0: expected str instance, int found
```
### ids that are dicts

https://sentry.io/open-data-services/cove/issues/160161169/

```
TypeError: unhashable type: 'dict'
(22 additional frame(s) were not displayed)
...
  File "cove/views.py", line 35, in wrapper
    return func(request, *args, **kwargs)
  File "cove/views.py", line 152, in explore
    validation_errors = common.get_schema_validation_errors(json_data, schema_url, schema_name, request.current_app, cell_source_map, heading_source_map) if schema_url else None
  File "cove/lib/common.py", line 186, in get_schema_validation_errors
    for n, e in enumerate(validator(schema, format_checker=format_checker, resolver=CustomRefResolver('', schema, schema_url=schema_url)).iter_errors(json_data)):
  File "cove/lib/common.py", line 33, in uniqueIds
    if item_id in all_ids:

TypeError: unhashable type: 'dict'
```

Bjwebb:
This is fixed in master now, and on the dev site http://dev.cove.opendataservices.coop/

PR #485, opened 2016-10-24
[#476] Don't 500 error on duplicate IDs with unexpected types

Bjwebb:
Adds some tests of this, and of the explore page more generally using
the hypothesis test library.
#476


Bjwebb:
This is fixed in master now, and on the dev site http://dev.cove.opendataservices.coop/

Issue #443, opened 2016-07-07
Roadmap runs until the end of July. What next?


caprenter:
Just a reminder that the roadmap document is due a review soon.

Bjwebb:
Links to the updated Cove roadmap https://github.com/OpenDataServices/cove/blob/master/ROADMAP.rst and the new flatten-tool roadmap https://github.com/OpenDataServices/flatten-tool/blob/master/ROADMAP.rst

PR #495, opened 2016-11-11
[#443] Belatedly add the roadmap from August

Bjwebb:
#443 

@caprenter could you double check this.


caprenter:
Reviewed and ready to go. Thanks @Bjwebb

PR #498, opened 2016-11-23
Update OCDS schema used to 1.0.2

Bjwebb:
https://github.com/open-contracting/standard/pull/402


robredpath:
@Bjwebb how much do we care about coveralls failing this for dropped coverage?

Bjwebb:
I don't think we do, actual coverage has not changed. In general coveralls is for guidance, rather than to be a definite blocker - I think the repo settings reflect that, GH will still let this be merged.

Issue #509, opened 2017-01-06
Move hosting & change the URL


robredpath:
**As a user, I want the Data Quality tool to look and feel like a 360Giving tool so that I have trust in it**

We'll put the new site on the new server at the new address, then update links.

We'll do this with 16.04, unless there are major issues. We'll use Let's Encrypt by default; same proviso.

caprenter:
Some links that will need updating:
http://www.threesixtygiving.org/support/resources/
http://www.threesixtygiving.org/data/publish-data/

Bjwebb:
We have a dev server set up, under the 360Giving's bytemark account, and using Ubuntu 16.04 -   http://cove.cove-360-dev.default.threesixtygiving.uk0.bigv.io/

Remaining tasks:
- [ ] Removing OCDS from this deploy
- [ ] Setting up certbot (Let's encrypt client) - @idlemoor is currently looking into this on another server, but we can then deploy the same salt config on the 360 cove servers
- [ ] Set up a live server (should be quick, using the salt config we've written for dev)
- [ ] Point core 360 cove subdomain at live server (@robredpath is asking what this should be)

Bjwebb:
Functional tests run successfully against the dev server with `BROWSER=PhantomJS CUSTOM_SERVER_URL=http://cove.cove-360-dev.default.threesixtygiving.uk0.bigv.io PREFIX_360=/ py.test fts` once I fixed a couple of bugs with them https://github.com/OpenDataServices/cove/commit/9e0cb6442e116f607231d691453bde70c977c27c

Bjwebb:
We have https: https://cove.cove-360-dev.default.threesixtygiving.uk0.bigv.io/

Automatic renewal is currently broken though, I need to fix some interaction between `.well-known` and `ProxyPass`

Bjwebb:
Dev server is now at https://master.cove-360-dev.default.threesixtygiving.uk0.bigv.io/ for more consistency in naming.
Live server is mostly set up, still need to point the subdomain at it though.

robredpath:
360 have approved  dataquality.threesixtygiving.org with a 301 redirect from the old URL, then a review using analytics to see if there are any links that need updating / when it's safe to switch off

Bjwebb:
https://dataquality.threesixtygiving.org/ is set up now.

PR #524, opened 2017-01-12
[#509] Make sure we use PREFIX_360 consistently in functional tests

Bjwebb:


Bjwebb:
This should be fine to merge with the unrelated coveralls oddness. I will separately see if I can fix that non-determinism today though.

Bjwebb:
https://github.com/OpenDataServices/cove/pull/536 removes the need for this.

Issue #523, opened 2017-01-11
Date is not in datetime format - error message can lead to confusion


caprenter:
We've had an issue reported where a user got the message:
"Date is not in datetime format" on a date that is permitted to be YYYY-MM-DD, but by following the link to the common errors page then turned them all in to datetimes.

We have a fudge in the schema validation that allows YYYY-MM-DD in datetime checks, AND we have an intercept message that we use to inform users of a problem.

To test this, use this fixture:
https://github.com/OpenDataServices/cove/blob/master/cove/fixtures/WellcomeTrust-grants_2_grants.xlsx?raw=true

I suggest we:
1) Change the error message to "Date is not in correct format"
2) Change the common errors page to make it clear that this may refer to a datetime or a date

For our use this links to an internal ticket no: 7185

ekoner:
Potential issues:

1. Wrong date i.e. dates that don't exist like 2017-02-31 - 31st of Feb 2017

2. Date format i.e. not using the right format of which there are currently a number supported - as described above by David.

3. Other issues with dates e.g. there's a text or some other non-date content.

Bjwebb:
This is fixed at https://dataquality.threesixtygiving.org/common_errors#date-time

PR #533, opened 2017-01-17
[#523] Provides alternate text for date errors

caprenter:
Changes the error message and the write up on the common errors page


Bjwebb:
@caprenter Caught by your own tests here I think ;)

caprenter:
Yep, and by committing as I ran out of the door.

Issue #559, opened 2017-01-31
Data Supplied box - "to" is in bold (between dates) 


stevieflow:
https://dataquality.threesixtygiving.org/data/398cdb32-bb27-40e7-8a28-c5cdef523caa

> 2013-07-07T00:00:00+00:00 to 2016-11-08T00:00:00+00:00.

In the above, the "to" is in bold.  It may be better to remove the bold.

caprenter:
I'm seeing this type of sentence:

    This file contains 69 grants from 1 funder to 67 recipients awarded from 2015-11-30T00:00:00+00:00 to 2016-02-06T00:00:00+00:00. 

Gooder Inglish would use 'between' and 'and' instead of from and to - IMHO

    This file contains 69 grants from 1 funder to 67 recipients awarded between 2015-11-30T00:00:00+00:00 and 2016-02-06T00:00:00+00:00.

stevieflow:
@caprenter good point - I'd support that.  And not in bold

edugomez:
@caprenter I will change that

PR #560, opened 2017-01-31
[#559] Un-bold 'to' in 360 explore Data Supplied in 'from date to date'

edugomez:
#559


Bjwebb:
If you rebase this onto the current master branch, then the coveralls issue should go away.

Issue #527, opened 2017-01-17
Support deprecation warnings


robredpath:
**As a user, I want to be warned when schema features that I’ve used are subject to deprecation warnings so that I can address them early.**

Deprecation is a 1.1 feature that still needs to be added to the schema. But, to support 1.1, we need to make CoVE pick up on deprecation warnings
Standard issue concerning how we deprecate fields: [OCDS #401](https://github.com/open-contracting/standard/issues/401)
Our belief is that it’s safe for us to go ahead and implement as per the issue; we recognise the risk that it may be different in the end

We're pretty certain that the way the schema is going to end up is as in the issue

robredpath:
UI is to add another box with a table in it, providing first 3 examples

robredpath:
Copy is programmatically pulled from the schema when it's ready

robredpath:
@edugomez this is in review - please can you provide instructions for how I (or anyone else) can review it?

edugomez:
@robredpath [PR 561](https://github.com/OpenDataServices/cove/pull/561) requires code review of the functions getting the deprecated fields in the schema and the uploaded data. The tests (unit tests, see comment in PR for frontend tests) pass with the given test schema and data, but it would nice to see if the functions may be missing some corner cases (or not so corner!). For example, @kindly noticed (personal communication) a problem with `patternProperties` that probably deserve a separate issue (unless `patternProperties` they go altogether, which would be nice then :) ).

Another useful thing would be to check that the test data and schema with deprecated fields are good and fit for the purpose. 

BTW, I think we really need a live server we all can ssh into and pull a version of the code to show others, although I understand prob this is not the best place to discuss that.

Bjwebb:
Here's a copy of the code running on our server, that's set up to use the fixture schema: http://527-support-deprecation-warnings.dev.cove.opendataservices.coop/ocds/
The sample data link on that page is a good example.

robredpath:
At show-and-tell we discussed how a user would know what 'deprecated' means in practice. We talked about how a tooltip could be useful, as well as a link out to more complete documentation.

Bjwebb:
Usage Count: 0 ?

Bjwebb:
I've fixed the Usage Count problem.

Issue #526, opened 2017-01-17
Support multiple versions of the schema


robredpath:
**As a user, I want the validator to support multiple MAJOR.MINOR versions of the schema so that I can test my data against the latest version of the schema during upgrades**

Allow use of 1.1 including before 1.1 is officially launched 
Autodetection of version, but we should expose UI to override if it’s wrong CoVE #494
We always use latest PATCH for each version
Version in the schema standard issue OCDS https://github.com/open-contracting/standard/issues/301

robredpath:
UI is a dropdown and a button in the headlines box, as this is where we already specify which schema we've validated against.

robredpath:
Dropdown options - auto-detect / 1.0 / 1.1

robredpath:
@edugomez is there any WIP on this that you can share?

edugomez:
@robredpath I will push the branch with the code I have in a bit. As for the interface, it will look like this:
![ocds_526](https://cloud.githubusercontent.com/assets/5618508/22549926/aa89d0a2-e945-11e6-994e-27ee13e6030e.png)

edugomez:
@robredpath obviously feel free to comment/ ask to modify that ^^

Bjwebb:
Here is a deploy of the work so far http://526-support-multiple-schema-versions.dev.cove.opendataservices.coop/ocds/

robredpath:
It would be nice to have a more obvious version number in the headlines, as well as the link to the schema.

timgdavies:
Great work all

Layout wise, I'm wondering if the 'Validate against other schema version' could become it's own box under 'Convert' as that would make a more balanced page? 

Suggest that the drop-down then also defaults to the currently selected schema version - to give additional visual indication of what was selected.

edugomez:
> Layout wise, I'm wondering if the 'Validate against other schema version' could become it's own box under 'Convert' as that would make a more balanced page?
I was thinking exactly the same, my only doubt being that reference to the schema version is in the headlines box where the new select form is. I think we could add some heading to the new box under 'Convert'  with something like 'Your data has been checked using schema version <a>x.x.x</a>..' and  the new form underneath?

timgdavies:
Will the schema selected affect the conversation as well? I believe it would.

In that case, would the schema select work well top-right, above the convert box?

robredpath:
@edugomez can you post a screenshot of where this ended up?

edugomez:
[...]

PR #561, opened 2017-01-31
527 support deprecation warnings

edugomez:
#527

This PR is missing the functional test checking the additions to the UI for files with deprecated fields. As @Bjwebb suggested, it should be possible to test passing a schema with deprecated fields using the `settings` fixture in pytest-django, but let's wait until [#526](https://github.com/OpenDataServices/cove/issues/526) is done and then see what's the best way to test that.


Bjwebb:
Note to myself when I review properly: I'm not 100% if this code is handling schema refs the way we want.

cove/lib/common.py
```diff
@@ -234,3 +234,98 @@ def get_schema_validation_errors(json_data, schema_url, schema_name, current_app
...
+        if current_path:
+            path = current_path + (prop,)
+        else:
+            path = (prop,)
+
+        if "deprecated" in value and path not in deprecated_paths:
+                deprecated_paths.append(path)
+
+        if value.get('type') == 'object':
+            _get_schema_deprecated_paths('', '', value, path, deprecated_paths)
```
Bjwebb: It's a matter of taste, but I'd rather see `None` passed than the empty string.

cove/lib/common.py
```diff
@@ -234,3 +234,98 @@ def get_schema_validation_errors(json_data, schema_url, schema_name, current_app
...
+            obj = jsonref.load(schema_file)
+
+    for prop, value in obj['properties'].items():
+        if current_path:
+            path = current_path + (prop,)
+        else:
+            path = (prop,)
+
+        if "deprecated" in value and path not in deprecated_paths:
+                deprecated_paths.append(path)
```
Bjwebb: I notice we're not yet getting the reason for the depreaction from the schema (`value["deprecated"]["description"]`), or the `deprecatedVersion`.

cove/lib/common.py
```diff
@@ -234,3 +234,98 @@ def get_schema_validation_errors(json_data, schema_url, schema_name, current_app
...
     return dict(validation_errors)
+
+
+def _get_schema_deprecated_paths(schema_name, schema_url, obj=None, current_path=(), deprecated_paths=None):
+    '''Get a list of deprecated paths (as tuples) in a schema'''
+    if deprecated_paths is None:
+        deprecated_paths = []
+
+    if schema_url and schema_url.startswith("http"):
+        obj = jsonref.loads(requests.get(schema_url + schema_name).text)
```
Bjwebb: I think we want to be using CustomJsonrefLoader here https://github.com/OpenDataServices/cove/blob/043f8e9cfeed32bd4c093d3080997a5bedb6e6d0/cove/lib/common.py#L101 (the class I link to should probably have a docstring explaining why!)

cove/lib/common.py
```diff
@@ -234,3 +234,98 @@ def get_schema_validation_errors(json_data, schema_url, schema_name, current_app
...
+                generic_paths[generic_key][path + (key,)] = value
+            else:
+                generic_paths[generic_key] = {path + (key,): value}
+
+    return generic_paths
+
+
+def get_json_data_deprecated_fields(schema_name, schema_url, json_data):
+    deprecated_schema_paths = _get_schema_deprecated_paths(schema_name, schema_url)
+    paths_in_data = _get_json_data_generic_paths(json_data)
```
Bjwebb: I'd be interested to know how much memory this uses, particularly since this gets called even if their are no deprecated paths in the schema.

cove/lib/common.py
```diff
@@ -234,3 +234,98 @@ def get_schema_validation_errors(json_data, schema_url, schema_name, current_app
...
+
+        if value.get('type') == 'object':
+            _get_schema_deprecated_paths('', '', value, path, deprecated_paths)
+        elif value.get('type') == 'array' and value['items'].get('properties'):
+            _get_schema_deprecated_paths('', '', value['items'], path, deprecated_paths)
+
+    return deprecated_paths
+
+
+def _get_json_data_generic_paths(json_data, path=(), generic_paths=None):
```
Bjwebb: I like function! Looks like it could be useful in future for other things.

cove/lib/common.py
edugomez: I'll go through the comment soon, I'm currently adding some changes with things I missed and @kindly and @timgdavies mentioned today, eg. showing deprecation version and description in UI

PR #570, opened 2017-02-06
526 support multiple schema versions

edugomez:
#526


cove/input/models.py
```diff
@@ -29,6 +29,9 @@ class SuppliedData(models.Model):
     modified = models.DateTimeField(auto_now=True, null=True)
     rendered = models.BooleanField(default=False)
 
+    # Last schema version applied to the stored data
+    schema_version = models.CharField(max_length=10, null=True, default='')
```
Bjwebb: Not sure why this has null=True, and defaults to the empty string. Could we have default=None?

cove/settings.py
```diff
@@ -86,6 +86,16 @@
                       'record': 'record-package-schema.json'},
         'default': None
     },
+    'schema_version': {  # Default schema version
+        'cove-ocds': '1.0.2',
+        'cove-360': None,
+        'default': None
+    },
+    'schema_version_choices': {
+        'cove-ocds': ('1.0.2', '1.0.1', '1.0.0'),  # Keep descending order when adding new versions
```
Bjwebb: I think having patch versions here is temporary, so that we can test this. Is that right?

I think for the end of this sprint we want 1.0 and the 1.1 dev version.
I'm not sure how best to implement it, but one way would be to have:
`SCHEMA_URL_OCDS=http://standard.open-contracting.org/{}/en/`
and the list of tuples as `('1.1-dev', '1.0')`.

We probably need to discuss this with @kindly

edugomez: Yes, this has been change in the latest code push to produce `'1.1-dev'` and `'1.0'` (using a slightly different approach for schema versions in setting.py)

cove/input/migrations/0007_supplied_data_schema_version.py
```diff
@@ -0,0 +1,24 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='supplieddata',
+            name='schema_version',
+            field=models.CharField(max_length=10),
+        ),
+        migrations.AlterField(
+            model_name='supplieddata',
+            name='rendered',
+            field=models.BooleanField(default=False),
```
kindly: Need to check if this causes any damage.  This is a migration that should have been run a while ago.

cove/views.py
```diff
@@ -102,6 +101,40 @@ def explore(request, pk):
...
         "created_date": data.created.strftime("%A, %d %B %Y"),
     }
+    
+    schema_name = request.cove_config['schema_name']
+    replace_conversion = False
+
+    if request.current_app == 'cove-ocds':
+        schema_version = data.schema_version or request.cove_config['schema_version']
+        schema_version_choices = request.cove_config['schema_version_choices']
+        schema_version_display_choices = sorted(list({display for display, version in schema_version_choices.values()}))
```
kindly: This makes no sense to me.  It is sorting the unique display names, but two display names could link to two different schemas.  It is the display names not the schemas that go into the database and that feels like it could go wrong.
Should we be storing 1.1 or 1.1-dev in the database? 
I also think it should be sorted by the config option order.

cove/settings.py
```diff
@@ -86,6 +85,21 @@
...
         'default': None
     },
+    'schema_version': {  # Default schema version
+        'cove-ocds': '1.0',
+        'cove-360': None,
+        'default': None
+    },
+    'schema_version_choices': {
+        # {version: (display, url)}
+        'cove-ocds': {
```
kindly: These options need some kind of way to decide the ordering when going in the dropdown list on the frontend.  So an ordered dict makes more sense.

I also do not understand this structure.  It says # version: (display, url) but display is used to go into the database not version.

cove/input/migrations/0007_supplied_data_schema_version.py
```diff
@@ -0,0 +1,24 @@
...
+    dependencies = [
+        ('input', '0006_supplieddata_rendered'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='supplieddata',
+            name='schema_version',
+            field=models.CharField(max_length=10),
+        ),
```
kindly: We may consider migrating old data to to have an explicit version, as all the old data will have nulls after migrating and the new ones will have a version.

cove/views.py
```diff
@@ -102,6 +101,40 @@ def explore(request, pk):
...
+            "data_uuid": pk,
+            "version_choices": schema_version_display_choices
+        })
+    else:
+        schema_url = request.cove_config['schema_url']
+
+    if 'version' in request.POST and request.POST['version'] != data.schema_version:
+        schema_user_choice = request.POST.get('version')
+        if schema_user_choice in schema_version_display_choices:
+            schema_version = schema_user_choice
```
kindly: That looks like the display version not the schema version.

cove/templates/explore_ocds_base.html
```diff
@@ -12,7 +12,34 @@
...
+      <br>
+      <strong>Validate and explore same data against a different version of the schema</strong>
+      <br>
+      <br>
+      <div>
+        <form class="form-inline" action="{% url 'cove:explore' data_uuid %}" method="post">
+        {% csrf_token %}
+          <select class="form-control" name="version">
+          {% for version in version_choices %}
+            <option value="{{ version }}" {% if version == version_used %}selected{% endif %}>
```
kindly: I think value should be the schema version not not the display version

kindly:
@Bjwebb if you could do a final check that would be good, but I am happy now.

cove/input/migrations/0007_supplied_data_schema_version.py
Bjwebb: Looks fine to me.

Bjwebb:
This looks good to go into master now, I'll hit merge once checks pass.

I've created a couple of issues around this that we might want to look at before this goes live https://github.com/OpenDataServices/cove/issues/582 https://github.com/OpenDataServices/cove/issues/584

Bjwebb:
Build currently fails, but it's not our fault: https://github.com/travis-ci/travis-ci/issues/7332
I'll restart the build once that is fixed.

Issue #597, opened 2017-03-01
Scottish charity numbers aren't recognised by additional checks


robredpath:
The charity number and recipient org ID check don't match valid Scottish charity numbers

PR #598, opened 2017-03-02
[#597] Allow scottish charity numbers and relax company check.

kindly:


Bjwebb:
Looks like this breaks support for Northern Irish charities atm.

cove/lib/threesixtygiving.py
```diff
@@ -131,13 +131,17 @@ def get_prefixes(distinct_identifiers):
 
 
 def check_charity_number(charity_number):
+    charity_number = str(charity_number)
+    if charity_number.startswith("SC"):
```
Bjwebb: Because this is called from OrganizationIdLooksInvalid.process, this would allow org ids of the form `GB-COH-SCxxxxxxx` when the correct org id would be `GB-SC-xxxxxxx`. A fix would be to strip these characters in RecipientOrgCharityNumber.process instead of here.

Relevant links:
http://www.threesixtygiving.org/standard/identifiers/#toc-organisation-identifier "2. Registered charity number (UK)"
http://org-id.guide/?coverage=GB&subnational=&structure=charity&sector=

Issue #579, opened 2017-02-10
Review memory usage of recent changes


Bjwebb:
In particular for depreaction, where I think we're making a new full representation of the data in memory: https://github.com/OpenDataServices/cove/pull/561#discussion_r99875902

edugomez:
@Bjwebb Is this urgent, as we have no control over the amount of data uploaded by user?

Bjwebb:
@edugomez I think it's a task for the next sprint.

kindly:
I have written a doc analysing the memory usage of COVE currently.

https://docs.google.com/document/d/1DwFGVLjqTD3yfWg-7zKGsl4X5L2B2R7yAJkGCyoBLUA/edit?usp=sharing

It can also be used as a guide to how we can do memory usage testing in future.

robredpath:
That's very interesting, thanks @kindly  - and thanks in particular for documenting your process so thoroughly, so we can investigate a bit more widely in the future.

jpmckinney:
Key conclusion:

> Deprecated fields take up around ~150MB of the ~500MB usage for this file so is a candidate for memory reduction.  However as just loading the JSON into memory takes ~240MB it will not save us anywhere near an order of magnitude to improve its memory usage.  We will have to careful not adding many more of these functions in future.

I think this issue can be closed, as there is no recommended action, besides continuing to consider memory usage in future.

PR #599, opened 2017-03-02
[#528] codelist verification.

kindly:
Looks up codelists from github and caches them.
Adds two new boxes to frontend one for closed and one for open
codelists.
Removes validation results for enums that also have codelists. This was
a bit messy as it was hard to get out correct properties from jsonschema
when dealing with arrays of enums.

### DIFF
--- a/cove/fixtures/tenders_releases_2_releases_codelists.json
+++ b/cove/fixtures/tenders_releases_2_releases_codelists.json
@@ -1,0 +1,115 @@
+{
+    "publisher": {
+        "scheme": null, 
+        "name": "Buyandsell.gc.ca", 
+        "uri": "https://buyandsell.gc.ca", 
+        "uid": null
+    }, 
+    "releases": [
+        {
+            "date": "2014-03-25T00:00:00.00Z", 
+            "language": "English", 
+            "id": "Buyandsell.gc.ca-2014-11-07-89f689cd-e784-4374-bb17-94144679d46f", 
+            "ocid": "PW-14-00627094", 
+            "initiationType": "tender", 
+            "tender": {
+                "documents": [
+                    {
+                        "url": "https://buyandsell.gc.ca/cds/public/2014/03/25/3a76d19e61825db7939e14afaa2002e3/14-2015_itt.pdf", 
+                        "id": "https://buyandsell.gc.ca/cds/public/2014/03/25/3a76d19e61825db7939e14afaa2002e3/14-2015_itt.pdf"
+                    }, 
+                    {
+                        "url": "https://buyandsell.gc.ca/cds/public/2014/03/25/a5d07c1166142b8b2f4183d7a5363fec/14-2015_itt_-_f.pdf", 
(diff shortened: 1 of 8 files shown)


cove/lib/common.py
```diff
@@ -342,3 +371,110 @@ def get_json_data_deprecated_fields(schema_name, schema_url, json_data):
...
+                title = record.get('Title') or record.get('Title_en')
+                codelists[codelist_file][code] = title
+
+        except requests.exceptions.RequestException:
+            raise
+
+    return codelists
+
+
+def _generate_data_path(json_data, path=()):
```
edugomez: This and the usage below is nice.

edugomez: I think we should use the same approach for deprecated paths in the data so as to avoid creating a fresh copy of the data in memory (thus addressing  issue #579)

cove/fixtures/tenders_releases_2_releases_codelists.json
```diff
@@ -0,0 +1,115 @@
...
+                    }
+                ], 
+                "tenderPeriod": {
+                    "endDate": "2014-04-22T00:00:00Z"
+                }, 
+                "items": [
+                    {
+                        "id": "1", 
+                        "classification": {
+                            "scheme": "GSINS",
```
kindly: That was added in order to make a codelist error as GSINS in not in the codelist file (but GSIN is), so is there on purpose.

cove/lib/common.py
```diff
@@ -158,13 +182,15 @@ def resolve_remote(self, uri):
...
+            result = super().resolve_remote(uri)
         else:
             with open(uri) as schema_file:
                 result = json.load(schema_file)
-            if self.cache_remote:
-                self.store[uri] = result
-            return result
+        
+        add_is_codelist(result)
+        self.store[uri] = result
```
Bjwebb: I'm not clear why `if self.cache_remote:` has been removed.

cove/lib/common.py
```diff
@@ -342,3 +371,112 @@ def get_json_data_deprecated_fields(schema_name, schema_url, json_data):
...
+
+
+def _generate_data_path(json_data, path=()):
+    for key, value in json_data.items():
+        if not value:
+            continue
+        if isinstance(value, list):
+            if isinstance(value[0], dict):
+                for num, item in enumerate(value):
+                    yield from _generate_data_path(item, path + (key, num))
```
Bjwebb: This approach will internal server error for bad JSON, e.g.

```
{"releases": [{}, "test"]}
```

cove/lib/common.py
```diff
@@ -342,3 +371,112 @@ def get_json_data_deprecated_fields(schema_name, schema_url, json_data):
...
+        path_no_num = tuple(key for key in path if isinstance(key, str))
+
+        if path_no_num not in codelist_schema_paths:
+            continue
+
+        codelist, isopen = codelist_schema_paths[path_no_num]
+
+        codelist_values = codelists.get(codelist)
+        if not codelist_values:
+            continue
```
Bjwebb: I'm not sure what codelists we're silently not checking here.

Bjwebb: This is because we don't support codelists for extensions yet. Skipping here means we at least don't fail.

Issue #601, opened 2017-03-09
App Separation for IATI


robredpath:

edugomez:
- Keep a single repo and break current `cove` into `cove` (core), `cove360`, `coveOCDS`, `coveIATI`.  
- Apps will have their own **_settings.py_**. In principle, setting files won't extend a base one,  so they will have some level of repetition, although they should be more simply (ie. we can ge rid of `COVE_CONFIG_BY_NAMESPACE`).  
- Apps will have its own template folder. Templates should live inside another folder with name of the app, making easy to refer to either own or `cove` core templates with same names (ie. _base.html_).  
- At the repo root level three _"manage.py"_ files: **_manage360.py_**, **_manageOCDS.py_** and **_manageIATI.py_**.  
- `urls` and `tests` to be distributed among new independent apps. No need for _urls_multi.py_ and the use of `PREFIX`.  
- Decide the distribution of `views`, management `commands`, `templatetags`, etc... between core `cove` and other dependant apps.

edugomez:
### The refactoring so far:
- The current structure: 4 apps,  `cove` (core), `cove_360`, `cove_ocds` and `cove_iati` (all depending on `cove`). 
- Each app has a _settings.py_ file. All apps inherit and extend  the settings in `cove`. 
- There is a single _manage.py_ at the root of the project that needs to be run with _--settings_ option: e.g. `python manage.py runserver --settings=cove_ocds.settings`.  
- Existing tests are distributed among the apps, and like _manage.py__,  _pytest_ needs to be passed the _--ds_ option to find the right settings: e.g. `py.test cove_360/ --ds=cove_360.settings` to run the tests in `cove_360`.This means that tests need to be run separately for each app, which can be useful for development, but we may also need some shell script to run all the test in the repo in one go when needed. 
- _cove-resource-project_ is gone. The only thing left is a _dataload_ folder in `cove` without models and with a migration file to run and delete the tables in the database. Once that migration is run on any existing instance of CoVE the whole folder can be deleted. 
- App-specific sass files are within the apps and depend on the bootstrap files in `cove` ( although still writing generated css files to _static_ in `cove`)

### TODO
- Get rid of _dataload_ in`cove`
- Separate  _fts_ tests among the appropriate apps
- hypothesis tests still in `cove` and not refactored according to the new project structure
- home page (input page now) for OCDS missing 'Using the Validator'
- Travis needs to be able to run the tests in all the apps
- _docs_ may need some review/changes
[...]

Issue #529, opened 2017-01-17
Add extension support


robredpath:
**As an analyst developing extensions for OCDS, I want to be able to provide both my extension and my data so that I can validate against the extension**

Extensions in the file standard issue - [OCDS #301](https://github.com/open-contracting/standard/issues/301#issuecomment-247793820)

We think it’s important that we explain very clearly to people what’s happened to create an extended schema before we tell them whether we think the data is valid or not
Non-registry extensions should come with a warning!

We're not going to support this in flatten-tool in order to keep things simple and stable! This means that titles and types won’t be handled as well as they would be if extended schema was supplied. This shouldn't affect most users, even when using extensions. 

@Bjwebb did a seperate CLI proof of concept https://github.com/Bjwebb/ocds_cli_validator

We will:
- Check that a supplied extension is valid against the JSON Schema meta-schema 
- Then validate the data against the extended schema

If we can't download the extension, we'll alert the user to the fact and allow the user to choose to validate against the un-extended schema using the dropdown

Bjwebb:
> We're not going to support this in flatten-tool in order to keep things simple and stable! This means that titles and types won’t be handled as well as they would be if extended schema was supplied. This shouldn't affect most users, even when using extensions.

Looks like this is true for some of the OCDS PPP extensions (because they use arrays).

Bjwebb:
In order to support the PPP usecase, we need to run flatten-tool with the extended schema. I think this is now a feasible option, because:
- we've done some work to fix flatten-tool for some of the extension schemas https://github.com/OpenDataServices/cove/issues/544
- most users of extensions in cove will either be 1) supplying JSON and not hitting flatten-tool OR 2) need flatten-tool to be run with the extensions

timgdavies:
Issues:

* [ ] How do we validate extensions?
* [ ] How do we report to users on the extensions they used? (e.g. do we pick up information from the extension.json to link to READMEs etc; do we flag extensions known from the registry or not)
* [ ] Do we check on compatibility and dependencies of extensions? 
* [ ] Reporting on extension failures: e.g. failed download; failed to apply; 
* [ ] Do we partially apply extensions, or just validate without extensions if one fails?

Bjwebb:
@edugomez  Can this issue be closed? (have all the tasks listed above been done?)

edugomez:
[...]

PR #606, opened 2017-03-13
529 add ocds extension support

edugomez:
#529 

- Extension support for OCDS schema (for release_package_schema only)
- New `SchemaMixin`, `SchemaOCDS(SchemaMixin)` and `Schema360(SchemaMixin)` to deal in a unified way with schema urls and names (ie. release_schema, release_package_schema...)  versions, extensions, fields, etc... These classes are thought as an early contribution towards #601 and non-sprint [#7216](https://opendataservices.plan.io/issues/7216)
- Code refactoring, with `explore` view refactored  into `explore`,  `explore_ocds` , `explore_360` and  `render_explore`. Again, this should help with  #601 and [#7216](https://opendataservices.plan.io/issues/7216)


cove/lib/common.py
```diff
@@ -1,26 +1,280 @@
...
+        self.extended_schema_url = None
+
+        if select_version:
+            try:
+                self.version_choices[select_version]
+            except KeyError:
+                select_version = None
+                self.invalid_version_argument = True
+                print('Not a valid value for `version` argument: using version in the release '
+                      'data or the default version if version is missing in the release data')
```
Bjwebb: Should there be a print statement here?

cove/lib/common.py
```diff
@@ -1,26 +1,280 @@
...
+                package_schema_obj = self.deref_schema(release_pkg_schema_str)
+                package_schema_obj['properties']['releases']['items'].update(deref_release_schema_obj)
+            else:
+                package_schema_obj = self.deref_schema(self.release_pkg_schema_str)
+        return package_schema_obj
+
+    def create_extended_release_schema_file(self, upload_dir, upload_url):
+        filepath = os.path.join(upload_dir, 'extended_release_schema.json')
+        if not self.extended or os.path.exists(filepath):
+            return
```
Bjwebb: Returning here means that:
* If I refresh the page `schema_ocds.extended_schema_url` is not set, so the link in the template is broken.
* The extended schema file is not updated if I change the version.

PR #612, opened 2017-03-22
601 app separation with iati

kindly:
#601


.travis.yml
```diff
@@ -18,7 +18,11 @@ install:
 
 script:
   - "flake8 ."
-  - "py.test --cov cove -n 2"
-  - "python manage.py migrate; python manage.py compilemessages; DEBUG=false ALLOWED_HOSTS=localhost python manage.py runserver & java -jar dist/vnu.jar http://localhost:8000/ http://localhost:8000/360/ http://localhost:8000/ocds/ 'http://localhost:8000/360/?source_url=https://github.com/OpenDataServices/cove/raw/master/cove/fixtures/WellcomeTrust-grants_2_grants.csv' 'http://localhost:8000/ocds/?source_url=https://github.com/OpenDataServices/cove/raw/master/cove/fixtures/tenders_releases_2_releases.json' 'http://localhost:8000/360/common_errors/'"
+  - "DJANGO_SETTINGS_MODULE=cove.settings py.test cove --cov --cov-report="
+  - "DJANGO_SETTINGS_MODULE=cove_360.settings py.test cove_360 --cov-append --cov --cov-report="
+  - "DJANGO_SETTINGS_MODULE=cove_ocds.settings py.test cove_ocds --cov-append --cov"
```
Bjwebb: Why do some of these have --cov-report and some don't?

Bjwebb: Oh, I get it, this stops the coverage report from being shown until the final run.

cove/templates/500.html
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html' %}
+{% extends app_base_template %}
```
Bjwebb: Should this be `request.current_app_base_template` for consistency with error.html etc.?

cove_ocds/tests_functional.py
```diff
@@ -455,62 +308,3 @@ def mockflatten(input_name, output_name, *args, **kwargs):
...
-    browser.get(server_url + prefix + 'data/0')
-    assert "We don't seem to be able to find the data you requested." in browser.find_element_by_tag_name('body').text
-    # Test for well formed UUID that doesn't identify any dataset that exists
-    browser.get(server_url + prefix + 'data/38e267ce-d395-46ba-acbf-2540cdd0c810')
-    assert "We don't seem to be able to find the data you requested." in browser.find_element_by_tag_name('body').text
-    assert '360 Giving' not in browser.find_element_by_tag_name('body').text
-    #363 - Tests there is padding round the 'go to home' button
-    success_button = browser.find_element_by_class_name('success-button')
-    assert success_button.value_of_css_property('padding-bottom') == '20px'
-
```
Bjwebb: Are `test_URL_invalid_dataset_request` and `test_500_error` now only tested for 360, whereas before they were also tested for OCDS?

cove_ocds/test_input.py
```diff
@@ -0,0 +1,87 @@
+import os
```
Bjwebb: As you might guess, as a DRY fan, I don't particularly like that this has just been copy and pasted.

kindly: yep, I tried to come up with a solution for not doing this but failed when I was rushing to get this done.  I really needed class based testing but it acted weirdly with fixtures and did not want to spend ages figuring it out.  Definitely something to fix for the future

cove_ocds/tests_functional.py
kindly: This was a mistake and I was supposed to put them in the OCDS tests too.

cove/templates/500.html
kindly: There was something weird about how the the 500 view is loaded which did not give me that variable, so I had to do it this way.

cove/input/templates/input/input.html
```diff
@@ -1,11 +1,10 @@
-{% extends 'base.html' %}
+{% extends request.current_app_base_template %}
 {% load bootstrap3 %}
 {% load i18n %}
 
 {% block content %}
-
+<p>{{ request_name }}</p>
```
Bjwebb: I don't think this is used.

kindly: Not sure how that got in there or what its for!

Bjwebb: A grep reveals nothing, so I'm going to remove it.

Issue #613, opened 2017-03-24
flatten_dict assumes that all lists are lists of dicts


ekoner:
Internal Server Error

Something went wrong with your request. This could be due to a bug in the application and/or the data you supplied.

This error has been automatically reported to our development team. If you get in touch, please reference this error as 19265c9235f8444cb905c00ddd755795.

Same file passes on: http://cove.opendataservices.coop/360/data/8ff6a19c-84df-4f18-bac1-790f3c8e22e9

robredpath:
Replicated. @Bjwebb - this doesn't seem to have shown up in Sentry!

Bjwebb:
https://github.com/OpenDataServices/cove/blob/54607ad/cove_360/lib/threesixtygiving.py#L449

Bjwebb:
@ekoner This should be fixed now.

PR #616, opened 2017-03-28
[#613] Handle lists of strings in flatten_dict

Bjwebb:
#613


Bjwebb:
We could test for things that look like emails in the new tests.

Bjwebb:
Actually, I think supporting lists of strings properly is more difficult, raising a new issue for this https://github.com/OpenDataServices/cove/issues/632
Merging this because it is at least an improvement.

PR #625, opened 2017-03-30
602 convert upload

kindly:


Bjwebb:
root_list_path should be `iati-activity`, see the example at https://flatten-tool.readthedocs.io/en/latest/usage-iati/

Bjwebb:
This is looking good, a couple of quick observations:
* I think it might be good to have a more explicit license/attribution for the bootswatch theme
* My cursor doesn't change to a pointer when I hover over the accordions.

Bjwebb:
Looking good. I think we should also be including this copyright file in our source tree https://github.com/thomaspark/bootswatch/blob/gh-pages/LICENSE

cove_360/sass/_bootstrap-variables-360.scss
```diff
@@ -1,6 +1,8 @@
 $bootstrap-sass-asset-helper: false !default;
 // Flatly 3.3.6
+// Bootswatch
 // Variables
+// MIT License (see LICENCE.srt)
```
Bjwebb: Very minor point, put this is a typo `srt` not `rst`

Bjwebb:
I've merged master into this branch so I can use it as the basis for IATI validation work. I've broken some OCDS functionality though. @edugomez if you're looking for something else to do this afternoon, could you have a go at fixing this please.

edugomez:
@Bjwebb  this now fails due to coveralls, tests are fixed. Is this good to go?

Issue #622, opened 2017-03-28
Return from create_extended_release_schema_file too early


Bjwebb:
https://github.com/OpenDataServices/cove/blob/39639d34993f0c96323c4b9e718fe1d5c4c0b298/cove_ocds/lib/schema.py#L139

Returning here means that:
* If I refresh the page `schema_ocds.extended_schema_url` is not set, so the link in the template is broken.
* The extended schema file is not updated if I change the version.

PR #628, opened 2017-03-31
[#622] Make sure exteded schema is remade.

kindly:


cove_ocds/views.py
```diff
@@ -103,32 +103,31 @@ def explore_ocds(request, pk):
...
-                    if os.path.exists(converted_path + '.xlsx'):
-                        replace = True
+                    replace = True
 
                 url = schema_ocds.release_schema_url
                 if schema_ocds.extensions:
                     schema_ocds.get_release_schema_obj()
                     if schema_ocds.extended:
-                        schema_ocds.create_extended_release_schema_file(db_data.upload_dir(), db_data.upload_url())
+                        schema_ocds.create_extended_release_schema_file(db_data.upload_dir(), db_data.upload_url(), replace=True)
```
Bjwebb: So, `create_extended_release_schema_file` is always called with `replace=True`?

cove_ocds/tests.py
```diff
@@ -620,3 +620,28 @@ def test_schema_ocds_extended_release_schema_file():
     assert not schema.extended
     assert not schema.extended_schema_file
     assert not schema.extended_schema_url
+
+
+@pytest.mark.django_db
+def test_schema_after_version_change(client):
```
Bjwebb: These tests look good, but they only test the second bullet point (not the first) on https://github.com/OpenDataServices/cove/issues/622

Issue #437, opened 2016-06-22
Separate translations by app


Bjwebb:
https://www.transifex.com/OpenDataServices/cove/translate/#es/core/89924590?translated=no

I don't think this is a blocker to this going live, since there's no real demand for Spanish for 360Giving atm.

Bjwebb:
Following app seperation https://github.com/OpenDataServices/cove/issues/601 there is a way to do this now.

PR #631, opened 2017-04-04
[#437] Separate translations by app

Bjwebb:


Bjwebb:
Tests pass on this now, so i think it's ready to merge once reviewed.

cove/locale/es/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,1035 @@
...
+#: cove/templates/500.html:14
+msgid ""
+"This error has been automatically reported to our development team. If you "
+msgstr ""
+"Este error se ha informado de forma automática a nuestro equipo de "
+"desarrollo. Si tu"
+
+#: cove/templates/500.html:14
+msgid "please reference this error as "
+msgstr "por favor haz referencia a este error como"
```
edugomez: The Spanish translation is a bit inconsistent as it mixes formal ('usted) and informal ('tú') manners of addressing users. I tried to follow suit and use the formal one as that is what I saw in previous translations, but as exemplified in this line it is not consistent.

Issue #641, opened 2017-04-12
Use flatten-tool's unflatten with metatab


edugomez:
Use recently added metatab features in flatten-tool so that data releases in flat format (csv or xlsx) can specify version, extensions, etc in a 'Meta' sheet 

(Related to ocds-show issue https://github.com/open-contracting/ocds-show/issues/12)

PR #647, opened 2017-04-20
641 use flatten tool metatab

edugomez:
#641


Bjwebb:
There's a deploy of this at http://641-use-flatten-tool-metatab.dev.cove.opendataservices.coop/validator/ now.

PR #653, opened 2017-04-25
Apply Django security update to requirements_dev also

Bjwebb:


Bjwebb:
Looks like this might be failing due to changes in the OCDS standard repo. @kindly do you know anything about this?

edugomez:
@Bjwebb the problem is this 404 at https://raw.githubusercontent.com/open-contracting/standard/1.1-dev/standard/schema/codelists/releatedProcessScheme.csv,  an url buit in Cove in `cove.lib.common._load_codelists` to retrieve a codelist file, notice the typo in _re**lea**tedProcessScheme.csv_. `_load_codelists` gets a `requests` exception and completely swallows it silently (perhaps not ideal behaviour for debugging).

Bjwebb:
Thanks @edugomez, the OCDS standard PR you made is merged now https://github.com/open-contracting/standard/pull/449.

Based on local testing, I think this will still fail, because we look for `procurementCategory`, but this is now called `mainProcurementCategory` - we just need to change that key.

Bjwebb:
@edugomez if you approve and merge this pull request, you should be able to use it as the base for other branches, so that they don't fail.

edugomez:
@Bjwebb that's cool, merging now

Issue #548, opened 2017-01-30
Add header link back to main 360Giving site


robredpath:

robredpath:
Internal issue #7918 t

PR #659, opened 2017-04-27
[#548] Add link to 360Giving home in heading tagline

edugomez:


cove/static/dataexplore/css/style.css
```diff
@@ -1,8 +1,12 @@
 .page-header a {
-  /*background: url(http://ocds.wpengine.com/wp-content/uploads/2014/07/logo_web3.png) no-repeat !important;*/
+  /* background: url(http://ocds.wpengine.com/wp-content/uploads/2014/07/logo_web3.png) no-repeat !important;*/
```
robredpath: We can probably remove this line, can't we?

Issue #639, opened 2017-04-11
360: Do people know that important data is via "Data supplied"


stevieflow:
When working with users, people tended to skip over the "data supplied" tab, as this gives the _impression_ that it's just a confirmation statement. 

On review, we know that the highest/lowest value and date is included in this.  This can be an important data quality indicator (particularly if there is an outlier date/value, perhaps indicating an input error).  

It could be useful for us to reconsider where we situate these statements - and/or relabel this box to "Data summary" (or similar)

robredpath:
On reflection, @stevieflow , it's a bit of an oddity that the summary data is alongside the Data Supplied statement in any case. 

How about another box that's non-collapsible to contain the Data Summary?

robredpath:
> That might in turn mean that the "Data Supplied" drop down is rendered meaningless?

I think it's important that the information about when a file was downloaded and from where be easily available, mostly for someone who's been sent a link to a set of results and wants to know if the data used was the most recent version. I have no particular attachment to where that information lives, though.

stevieflow:
Agreed.  Sorry - I mean that the *drop down* might be meaningless function, if it is just one statement to be revealed

robredpath:
I like the fact that the results page is very clean, and almost all the details are hidden until you want to see them. One of the bits of feedback about the 'old' CoVE was that it was very overwhelming with lots of information, and it was hard to see where your attention should go. The new single-column layout may help a lot with that, so hiding details as well might not be necessary.

stevieflow:
> The new single-column layout 

Is this a new new thing, or the current new thing?

robredpath:
The current new thing. What you see at dataquality.threesixtygiving.org is the single-column layout I'm referring to and have no current plans to change

edugomez:
@robredpath @stevieflow Would just changing the name of the section from _Data Supplied_ to _Data Summary_ suffice to fix this issue? Or @stevieflow, would you scrap the section and move the info to Additional Checks? I personally prefer the former (renaming section)

stevieflow:
> Would just changing the name of the section from Data Supplied to Data Summary suffice to fix this issue?

Agree.  Renaming the header seems like less pain, and we can then review

PR #663, opened 2017-04-28
[#639] cove_360: rename Data Supplied section as Data Summary

edugomez:


stevieflow:
Checking the diff looks good, but will leave to @robredpath to fully review.  Thanks

Issue #201, opened 2015-12-15
360Giving: check for Title length 


caprenter:
The title must not be more than 140 characters according to: http://www.threesixtygiving.org/standard/reference/

It would be good if CoVE could flag titles longer than 140chars.

This issue is copied over from:
https://github.com/ThreeSixtyGiving/standard/issues/97
where altering the schema to enable this check is discussed

stevieflow:
Is this an example of additional tests (or is it a part of schema tests?)?

robredpath:
Note that the standard says "should", not "must"

robredpath:
This sprint: research what our approach to warning systems / audits / extra tests / 'should's can be

Bjwebb:
We now have the additional checks framework in place such that we could do this.

PR #667, opened 2017-05-02
[#201] Additional checks 360: check for grant title length < 140 chars

edugomez:


cove_360/lib/threesixtygiving.py
```diff
@@ -396,6 +396,18 @@ def process(self, grant, grant_flat, path_prefix):
...
 
+class TitleLength(AdditionalTest):
+    def process(self, grant, grant_flat, path_prefix):
+        title = grant.get("title", '')
+        if len(title) > 140:
+            self.failed = True
+            self.count += 1
+            self.json_locations.append(path_prefix + '/title')
+
+        self.heading = self.format_heading_count("a title longer that recommended")
```
robredpath: longer than*

cove_360/tests.py
```diff
@@ -352,12 +353,16 @@
...
-    ({'heading': "2 grants have a title and a description that are the same",
+    ({'heading': "1 grant has a title and a description that are the same",
       'message': ("Users may find that the data is less useful as they are unable to "
                   "discover more about the grants. Consider including a more detailed "
                   "description if you have one.")},
-     ['grants/1/description', 'grants/2/description'],
-     [['grants', 'Z', 3, 'Description'], ['grants', 'Z', 4, 'Description']]),
+     ['grants/2/description'],
+     [['grants', 'Z', 4, 'Description']]),
+    ({'heading': "1 grant has a title longer that recommended",
```
robredpath: longer than*

cove_360/lib/threesixtygiving.py
```diff
@@ -396,6 +396,18 @@ def process(self, grant, grant_flat, path_prefix):
         self.message = "Users may find that the data is less useful as they are unable to discover more about the grants. Consider including a more detailed description if you have one."
 
 
+class TitleLength(AdditionalTest):
+    def process(self, grant, grant_flat, path_prefix):
+        title = grant.get("title", '')
```
robredpath: How does this handle the case of (invalid) data that doesn't have a title field?

robredpath: To be clear: in the current state, we don't expect Additional Checks to produce meaningful results for invalid data. But I do want to check that it doesn't break horribly (eg throw a 500) if the data is invalid.

cove_360/lib/threesixtygiving.py
edugomez: Will change that

cove_360/lib/threesixtygiving.py
edugomez: If there's no title field this test won't produce any output (`grant.get('title', '')` will avoid any `KeyError` raising a 500 in the server if there's no title field)

PR #691, opened 2017-05-26
688 - Missing ocds deprecated fields

edugomez:


cove/lib/common.py
```diff
@@ -350,14 +350,15 @@ def _get_json_data_generic_paths(json_data, path=(), generic_paths=None):
             generic_paths[path] = []
 
     for key, value in iterable:
+        generic_key = tuple(i for i in path + (key,) if type(i) != int)
```
edugomez: @kindly, sorry I had no time to convert this into a generator (a bit tricky) but will do at some point as part of a ticket (8027) on the non- sprint board in Planio.

Issue #734, opened 2017-07-02
OCDS Validator: Reports tender.id missing when tender.procuringEntity.id missing


duncandewhurst:
When validating against OCDS 1.1 the validator reports that tender.id is missing when tender.procuringEntity.id is missing (and tender.id is present). See [example](http://standard.open-contracting.org/validator/data/8825cf1a-8a5a-4a2d-870a-f8731266b194):

![image](https://user-images.githubusercontent.com/19265814/27771701-0bcef708-5f4c-11e7-9ce8-c358c9ba6044.png)

Bjwebb:
This is fixed on dev now, you can test it at http://release-201707.dev.cove.opendataservices.coop/validator/

PR #735, opened 2017-07-04
[#734] fix field name logic

kindly:
#734 

Logic for describing the field name in validation error results is wrong and too the grandparent instead of the parent.


Bjwebb:
We discussed this briefly this morning. Would be good to have some tests of the correct behaviour.

PR #738, opened 2017-07-04
Review and fix Spanish translations

edugomez:


edugomez:
@Bjwebb could you approve this to get it out of the way?

Issue #707, opened 2017-06-14
OCDS Validator: OCDS 1.1 records including the version key don't validate


duncandewhurst:
When trying to validate [this sample OCDS 1.1 record](https://gist.githubusercontent.com/duncandewhurst/a9d22ef19f29dca1910ede9bdb4fb48d/raw/1ee452a43def064fc2644a70dd0d8eb46d27e43b/ocds-213czf-000-00001.json) with the version key set to 1.1 the validator returns [this error](http://standard.open-contracting.org/validator/data/c8340a77-54ae-4af4-a924-d26990be5662): ```'releases' is missing but required```

The same file validates with the version key [set to 1.0](http://standard.open-contracting.org/validator/data/39c97699-55e4-47b9-bec8-662f2ddb17cc) or [omitted](http://standard.open-contracting.org/validator/data/192b814c-af03-4ed0-b097-52b3031605bf), including manully changing the schema selector in the validator to 1.1.

Bjwebb:
Looks like this is partly a schema bug - "releases" is marked as required in the OCDS 1.1 record package schema https://github.com/open-contracting/standard/blob/1.1-dev/standard/schema/record-package-schema.json#L118

Bjwebb:
>  including manually changing the schema selector in the validator to 1.1

Looks like this is the bit that is broken in CoVE, the 1.0 record schema is used.

CoVE actually does the "correct" thing when the version key is set, but the schema is pretty obviously wrong, so I've opened a standard issue - https://github.com/open-contracting/standard/issues/479
Currently cove pulls from the frozen point release schema urls, which would require a 1.1.1 release to fix this.

Bjwebb:
We've fixed what we can in CoVE, but should add tests.

robredpath:
@Bjwebb I'm just reviewing OCDS CoVE issues - does there need to be a standard issue for this?

Bjwebb:
There is a standard issue: https://github.com/open-contracting/standard/issues/479

PR #748, opened 2017-07-11
Test that version switching happens correctly for record files

Bjwebb:
#707


cove_ocds/tests.py
```diff
@@ -674,6 +674,42 @@ def test_schema_after_version_change(client):
...
+
+    with open(os.path.join(data.upload_dir(), 'validation_errors-2.json')) as validation_errors_fp:
+        assert "'version' is missing but required" in validation_errors_fp.read()
+
+    resp = client.post(data.get_absolute_url(), {'version': '1.0'})
+    assert resp.status_code == 200
+
+    #with open(os.path.join(data.upload_dir(), 'extended_record_schema.json')) as extended_record_fp:
+    #    assert "mainProcurementCategory" not in json.load(extended_record_fp)['definitions']['Tender']['properties']
+
```
edugomez: @Bjwebb was this left here commented out intentionally ?

Bjwebb: Yes, to highlight that we might want to test something like this for records, but at the moment it doesn't happen - https://github.com/OpenDataServices/cove/issues/747. Perhaps I should add a note with a link to that issue.

edugomez: > Perhaps I should add a note with a link to that issue.

@Bjwebb I'm doing that right now

PR #773, opened 2017-07-24
Add a docker file

tobybatch:
I'd like to move the docker file into your CoVE repo, it seems to make more sense to have it there.  Is that OK?

It should work with other coves if you start it with a -e DJANGO_SETTINGS_MODULE=cove_iati.settings


Bjwebb:
I've been sent a more up to date Dockerfile over email, which I've replied with some queries about.

PR #793, opened 2017-07-28
Merge translations made for release into master

Bjwebb:


cove/locale/es/LC_MESSAGES/django.po
```diff
@@ -1032,3 +1037,139 @@ msgstr ""
...
+msgid "downloaded from "
+msgstr "Descargado desde"
+
+#: cove_iati/templates/cove_iati/explore.html:33
+msgid "uploaded "
+msgstr "Subido "
+
+#: cove_iati/templates/cove_iati/explore.html:35
+msgid "on "
+msgstr "en"
```
edugomez: @Bjwebb I realise that the right translation for this _on_  in this context is probably _el_. Can we change it just here or do we need to go to transifex, change and pull it again? I guess I need to do it in transifex otherwise _en_ will be pulled again next time around? Or we can just leave like this too, no big deal :)

edugomez: @Bjwebb I just changed this in transifex for cove and cove_ocds plus a couple of mismatched capitalisations

Bjwebb: Change on transifex and pull again.

PR #807, opened 2017-08-07
IATI ruleset first iteration

edugomez:
This PR is pretty much WIP and not meant to be merged yet:
- Not tested as the code will change substantially in the next iteration
- We need some changes in [bdd-tester](https://github.com/pwyf/bdd-tester) particularly in the output the library produces. The idea is to fork the repo and submit PRs to it.
- Paths for rule errors are xpath (e.g. no 0 indexes, [i] for array positions, etc..). In `cove_iati` We have the code to transform them into the shape we use everywhere else but it needs to be refactor to make it reusable.
- Some rules doesn't seem to work, e.g. not valid iso dates
- IATI rulesets for 1.x and 2.x are not clearly separated.


requirements_iati.txt
```diff
@@ -1,3 +1,4 @@
 -r requirements.txt
+-e git+https://github.com/pwyf/bdd-tester.git#egg=bdd-tester
```
Bjwebb: Can we pin this to a specific commit?

cove_iati/rulesets/iati_standard_ruleset/mandatory_elements.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Mandatory elements
+
+  Scenario Outline: activity-date @type="start-planned" or @type="start-actual" must be present (IATI 1.x)
```
Bjwebb: Can we remove the three `(IATI 1.x)` rules?

requirements_iati.txt
edugomez: @Bjwebb this change is now in its own commit ae819f8d531f592702086d7684fce16c23072fd4

Bjwebb: Sorry, what I mean is to specify the commit of bdd-tester to use, like we do for flattentool: https://github.com/OpenDataServices/cove/blob/master/requirements.txt#L3

Bjwebb:
Discussed on Tuesday that this is now at an appropriate point to merge for IATI CoVE (which is still only in dev, with some caveats).

Issue #802, opened 2017-08-04
360 CoVE: additional checks dont load for one publisher


stevieflow:
CiN - data loads, passes and converts, but not any additional checks:

https://dataquality.threesixtygiving.org/data/8697ec8f-6373-4387-8472-ce9d8b3a55bd

Bjwebb:
This should be fixed on http://release-201707.cove-360-dev.default.threesixtygiving.uk0.bigv.io/. If that now works as expected we can deploy the fix to the live site.

stevieflow:
Yep  - I got it to work too: http://release-201707.cove-360-dev.default.threesixtygiving.uk0.bigv.io/data/38923dfe-7ca0-45d1-af62-12f46d0c9789

Issue #801, opened 2017-08-04
360 CoVE aditional check : Incomplete recipient location data - wrong?


stevieflow:
(also linked to observations in https://github.com/OpenDataServices/cove/issues/800 )

This additional check:

> rants have incomplete recipient organisation information
> Your data is missing Recipient Org: Postal Code, Recipient Org: Location:Geographic Code or Recipient Org: Location: Geographic Code Type. Knowing the geographic location of recipient organisations allows users of your data to understand your data and combine it with other data sets to form a broader picture of grant-making.

Present in this test of JSON data, which includes postalCode for recipient orgs: https://dataquality.threesixtygiving.org/data/60485984-1b96-434f-9fc2-2722f15841da

Present in this test of spreadsheet data, which include Recipient Org:Postal Code: 
https://dataquality.threesixtygiving.org/data/28934e3b-3ea0-46cf-a7d0-f6f13383794d

cc/ @KDuerden

KDuerden:
I was curious to see message for data with the geocode & code type but no postcode so tried to load BBC CiN data. I'm getting validation error that means the additional aren't running, but the data does pass validation: https://dataquality.threesixtygiving.org/data/c843d643-fc2e-4449-8f9c-3daeef1bf66e

stevieflow:
@KDuerden I saw that too.  I'll create a new ticket for that, as it's not related to this specific issue.  Thanks

Bjwebb:
Currently the check is that all three of the fields must exist:
* Recipient Org:Postal Code
* Recipient Org:Location:Geographic Code
* Recipient Org:Location:Geographic Code Type

Is this correct?

KDuerden:
My understanding is this is supposed to highlight useful recipient org data
about location, so either Postcode OR Geocode + code type would provide
that. Suggesting an issue if both types of location data are not present
seems like overkill.

stevieflow:
Testing Esmee JSON data

https://esmeefairbairn.org.uk/userfiles/Documents/JSON%20grants%20list/Esmee_Grants_Since_2008.json

2907 grants

This contains a postalCode for a recipientOrganization

However, the CoVE test - http://release-201707.cove-360-dev.default.threesixtygiving.uk0.bigv.io/data/3a3c3fc4-db9a-4a6c-816f-38069e590244 -  still suggests:

> 2907 grants have incomplete recipient organisation information

edugomez:
[...]

Issue #800, opened 2017-08-04
360 CoVE: Additional check for Classifications: Title is wrong?


stevieflow:
We seem to be checking for ``Classification: Title`` whereas the schema has ``Classifications: Title``

This JSON file does have classification in, but CoVE states null: https://dataquality.threesixtygiving.org/data/60485984-1b96-434f-9fc2-2722f15841da

cc/ @KDuerden

stevieflow:
NB: I was trying to establish whether this was a JSON specific issue or more widespread. @KDuerden - know of an spreadsheet example that includes classifications?

@Bjwebb any reason why it should be JSON specific

KDuerden:
Quixote: https://dataquality.threesixtygiving.org/data/ae5bbf1e-3a7c-4762-9f25-02a46423ce56
http://grantnav.threesixtygiving.org/grant/360G-quixote-0063300000kufN9AAI
I first spotted this error when demoing with test data spreadsheet that included the Classifications:Title field (and postcode).

Bjwebb:
This should be fixed on http://release-201707.cove-360-dev.default.threesixtygiving.uk0.bigv.io/. If that now works as expected we can deploy the fix to the live site.

stevieflow:
I do not see that test in the same JSON anymore, so think it's good: 

http://release-201707.cove-360-dev.default.threesixtygiving.uk0.bigv.io/data/3a3c3fc4-db9a-4a6c-816f-38069e590244

PR #811, opened 2017-08-09
Fix 360Giving additional checks

Bjwebb:
#800 #802 #801


cove_360/lib/threesixtygiving.py
```diff
@@ -389,7 +389,7 @@ def process(self, grant, path_prefix):
             count_failure = False
             for num, organization in enumerate(grant['recipientOrganization']):
                 has_postal_code = organization.get('postalCode')
-                has_location_data = organization.get('location') and organization.get('location').get('geoCode') and organization.get('location').get('geoCodeType')
+                has_location_data = organization.get('location') and any(location.get('geoCode') and location.get('geoCodeType') for location in organization.get('location'))
                 complete_recipient_org_data = has_postal_code and has_location_data
```
edugomez: I think that making this `has_postal_code or has_location_data` instead would also address https://github.com/OpenDataServices/cove/issues/801

`self.message` in line 405 may need a minor tweak too:
https://github.com/OpenDataServices/cove/blob/3ba413119b4e43de2fdcff51628b5f59e0d2a942/cove_360/lib/threesixtygiving.py#L405

Bjwebb: Yes, that's right? Do you want to make these changes to the `800-802-fix-additional-checks` branch?

edugomez: Yes, I will do

Issue #655, opened 2017-04-26
Additional Checks section:  make counts clickable to see all instances


edugomez:
Derived from now close https://github.com/OpenDataServices/cove/issues/549

stevieflow:
#360 looks specifically at the counts clickable

stevieflow:
(as in issue #360 , not the 360 in general!)

Bjwebb:
This is now on dev: http://master.cove-360-dev.default.threesixtygiving.uk0.bigv.io/

PR #813, opened 2017-08-14
Additional checks 360 'See all' modal

edugomez:
#655

### DIFF
--- a/cove_360/tests.py
+++ b/cove_360/tests.py
@@ -109,6 +109,7 @@ GRANTS = {
                 'title': 'Exceptional and Extraordinary: unruly bodies and minds '
                          'in the medical museum. '}]}
 
+
 SOURCE_MAP = {
     'grants/0': [['grants', 2]],
     'grants/0/Co-applicant(s)': [['grants', 'E', 2, 'Co-applicant(s)']],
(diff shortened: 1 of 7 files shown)


cove_360/tests.py
```diff
@@ -286,21 +287,21 @@
                   "on why these grants are £0 might be useful to anyone using the data, "
                   "so consider adding an explanation to the description of the grant.")},
      ['grants/0/amountAwarded'],
-     [['grants', 'Q', 2, 'Amount Awarded']]),
+     [{'sheet': 'grants', 'letter': 'Q', 'row_number': 2, 'header': 'Amount Awarded'}]),
```
edugomez: This has been changed from list of lists to list of dicts to be in agreement with the data structure for validation spreadsheet source path for schema validation

edugomez:
360_see_all

Clicking 'See all' triggers the modal dialog with the location of all non-passing checks (only for >3)

robredpath:
I'd expect the 'see all' link to be after the 'first 3 locations' - is that possible? After the prose it feels like it's going to tell me more about what the check means rather than where they are in my data.

edugomez:
> I'd expect the 'see all' link to be after the 'first 3 locations' - is that possible?
 
Yes, that makes sense.

edugomez:
360_see_all

360_modal

edugomez:
@Bjwebb please review when you have a minute, I'd like to move to "done" in the non-sprint board.

Bjwebb:
There's no "see all" link for spreadsheets, is this correct?

edugomez:
> There's no "see all" link for spreadsheets, is this correct?

No, it should be there.

PR #820, opened 2017-08-22
Avoid 500 for non existing ruleset directory

edugomez:


edugomez:
@Bjwebb great, thanks!

edugomez:
@Bjwebb for some reason this is still failing

Issue #823, opened 2017-08-31
OCDS Validator: 500 error when validating bids extension


duncandewhurst:
Trying to validate a file with the bids extension declared in the extensions array results in a [500 error](http://standard.open-contracting.org/validator/data/87e6e6ac-1578-45ad-9ee2-96801cd829b7)

Originally noted on [this file](https://gist.github.com/duncandewhurst/459078c48b924c0f8df5d72b1206d28d) from OpenNWT.

Reproduced when trying to validate the following data (taken from the [extensions](http://standard.open-contracting.org/latest/en/extensions/#using-extensions) section of the docs)

```json
{
    "extensions":["https://raw.githubusercontent.com/open-contracting/ocds_bid_extension/v1.1/extension.json"],
    "releases":[]
}
```

edugomez:
The problem seems to be this 
```jsonref.JsonRefError: ("Unresolvable JSON pointer: '/definitions/RequirementResponse'", OrderedDict([('$ref', '#/definitions/RequirementResponse')]))```

The definition of ```RequirementResponse```is in [ocds_requirements_extension](https://github.com/open-contracting/ocds_requirements_extension).

So adding _"https://raw.githubusercontent.com/open-contracting/ocds_requirements_extension/master/extension.json"_ to the extension array fixes the problem.

edugomez:
BTW, good reminder that (atm) we deal badly with extension dependencies.

edugomez:
Relate to https://github.com/OpenDataServices/cove/issues/809

duncandewhurst:
thanks @edugomez - since this is a community extension I think it's fine to use `master`, the `ppp` release was created so we have fixed version for the ppp profile (to avoid breaking translations)

@dwasyl - see https://github.com/OpenDataServices/cove/issues/823#issuecomment-326273195 for the solution to the 500 error you are getting

edugomez:
Closing this one, but the issue of dealing more gracefully with missing JSON pointers for jsonref is still outstanding at https://github.com/OpenDataServices/cove/issues/809

edugomez:
FYI,  the problem is related to the discussion in this issue in OCDS extension registry https://github.com/open-contracting/extension_registry/issues/56, also in this one https://github.com/open-contracting/extension_registry/issues/47

Issue #809, opened 2017-08-08
Validating extension against 1.0 displays "Internal Server Error"


julijahansen:
When you try to validate a file with extensions in the validator tool against 1.0 it gives the following error instead of a warning message:
![image](https://user-images.githubusercontent.com/30236091/29070233-55f1f144-7c36-11e7-8d17-a050fd4d18e4.png)

Could we fix it so it gives e.g. a warning mesage instead of showing Internal Server Error?

When you try to vaidate the same file against 1.1 it does work:
![image](https://user-images.githubusercontent.com/30236091/29070264-78e06eec-7c36-11e7-8a28-4cc16de36c3b.png)

Validator results are [here](http://standard.open-contracting.org/validator/data/58538b01-5ef8-472a-8c79-8148992c3c40 )
And sample JSON file is attached.
[sample_ext.txt](https://github.com/OpenDataServices/cove/files/1207860/sample_ext.txt)

Bjwebb:
The underlying error is:
```
jsonref.JsonRefError: ("Unresolvable JSON pointer: '/definitions/OrganizationReference'", OrderedDict([('$ref', '#/definitions/OrganizationReference'), ('title', 'Related tenderer'), ('description', 'Where this requirement response relates to a tenderer and is provided by the buyer or procuring entity this field should be used to reference the entry in the parties section for the tenderer the response relates to.')]))
```
This is because the extension references `#/definitions/OrganizationReference` which exists in the 1.1 schema, but not the 1.0 schema. Other extensions that don't do this should work fine.

But yes, CoVE should handle this kind of failure more gracefully.

Bjwebb:
This should be fixed on dev: http://dev.cove.opendataservices.coop/validator/

PR #826, opened 2017-09-01
Handle jsonref exceptions when resolving refs in SchemaOCDS

edugomez:
https://github.com/OpenDataServices/cove/issues/809 and also (partially, as it is also a problem with extension dependencies) the cause of a 500 response in https://github.com/OpenDataServices/cove/issues/823

### DIFF
--- a/cove/lib/common.py
+++ b/cove/lib/common.py
@@ -93,8 +93,18 @@ class SchemaJsonMixin():
         return json.loads(self.release_pkg_schema_str)
 
     def deref_schema(self, schema_str):
-        return jsonref.loads(schema_str, loader=CustomJsonrefLoader(schema_url=self.schema_host),
-                             object_pairs_hook=OrderedDict)
+        try:
+            deref_obj = jsonref.loads(
+                schema_str,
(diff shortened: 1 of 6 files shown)


cove/lib/common.py
```diff
@@ -93,8 +93,15 @@ def _release_pkg_schema_obj(self):
         return json.loads(self.release_pkg_schema_str)
 
     def deref_schema(self, schema_str):
-        return jsonref.loads(schema_str, loader=CustomJsonrefLoader(schema_url=self.schema_host),
-                             object_pairs_hook=OrderedDict)
+        loader = CustomJsonrefLoader(schema_url=self.schema_host)
+        try:
+            deref_obj = jsonref.loads(schema_str, loader=loader, object_pairs_hook=OrderedDict)
+            # Force evaluation of jsonref.loads here
+            repr(deref_obj)
```
edugomez: I found quite difficult to force evaluation of the `deref_object` here, e.g `dict(jsonref.loads(schema_str ...) )` doesn't seem to work. I saw in the source code for JsonRef that it evaluates on representation, so this works, although I wonder how correct this approach is.

Issue #822, opened 2017-08-30
CoVE should explicitly report in prose the precise schema it is using for validation


jpmckinney:
From https://crm.open-contracting.org/issues/2427#note-33

The feedback from the validator (e.g. http://standard.open-contracting.org/validator/data/b71de36f-90b5-4e6e-a4bf-26e610f6244a) is unclear as to which schema it is validating against. There's a reference to 'schema version 1.0' (although it's actually 1.0.3), and if you click the link you can tell from the URL that it's validating against the release package schema, but this should be made clear in prose.

It also doesn't seem possible for me as a user to select which schema to validate against. As it happens, the data does validate against the release schema. It's unclear to me why the release package schema was selected, but at any rate it would have been more obvious what was going on if this selection were reported in prose instead of having to inspect a link's URL.

edugomez:
> CoVE should explicitly report in prose the precise schema it is using for validation

It does: 

_This data has been **checked against OCDS schema version 1.1.** You can choose a different version of the schema to validate and explore your data._ 

> There's a reference to 'schema version 1.0' (although it's actually 1.0.3)

This is intentional, users can choose major.minor but not patch (we map and apply the latest patch). This is agreement with ```version``` spec in the _release-package-schema.json_: 
```
       "version": {
            "title": "OCDS schema version",
            "description": "The version of the OCDS schema used in this package, expressed as major.minor For example: 1.0 or 1.1",
            "type": "string",
            "pattern": "^(\\d+\\.)(\\d+)$"
        },
```

> It also doesn't seem possible for me as a user to select which schema to validate against

_"Validate and explore same data against a different version of the schema"_ in the Schema box.

jpmckinney:
When I saw 'precise schema' I mean explicitly reporting which of these is being used:

* release schema
* release package schema
* record package schema
* versioned release validation schema

This is not reported as far as I can see.

edugomez:
Ah, OK. I don't think Cove should allow users to do that. Data is meant to be published either as a release-package or a record-package. Cove detects whether you are submitting a release-package or a record-package and validates accordingly. If you want to validate a single release, you should upload a release-package with a single release in it (or a record-package with a single release covering the whole contracting process, although I don't know if that is something occurring in real life publishing of OCDS data)

I think that uploading a release-package and then asking to validate against the record schema wouldn't be quite right.

[...]

PR #831, opened 2017-09-11
Error message for missing OCDS package format

edugomez:
#822 
@julijahansen  @timgdavies could you please check this PR? Please see pictures below.


edugomez:
missing_pckg_a

This is the error message when a user tries to upload data without a ```releases``` or ```records``` field in it,.e.g. when it tries to validate data following the release-schema outside a package. This is for **JSON data only**.

edugomez:
missing_pckg_b

Explicit mention to the package schema used to validate the data.

timgdavies:
The second screenshot looks great. For the first one, perhaps the following language:

"We could not detect a package structure at the top-level of your data. OCDS releases and records should be published within a release or record package to provides important meta-data. For more information, please refer to the Releases and Records section in the OCDS documentation." 

(with the links as you had them) 

Otherwise - looks good.

PR #840, opened 2017-09-19
643 describe additional checks

robredpath:


stevieflow:
thanks @robredpath - just to double check, is this adding the page, but awaiting content, or inclusive of?

robredpath:
This PR includes all of the content. It pulls the description of the check and the text used in the check directly from the code so that there's no doubt over its provenance. @stevieflow

stevieflow:
OK thanks @robredpath 

If @KDuerden is to review the text (via gdoc), then is there merit in accepting this PR, with the knowledge that the content *will* change.  I have no opinion - just checking

robredpath:
@stevieflow this page is currently designed for the benefit for 360 and ODSC. It won't be linked to from anywhere as a result of this PR. It's just so that when someone asks "what are all the additional checks that currently get run, and how do they work?", we can give an authoritative answer *for the version of CoVE that they're looking at*

stevieflow:
Excellent, thanks @robredpath

robredpath:
@Bjwebb yup, added

cove_360/templates/cove_360/additional_checks.html
```diff
@@ -0,0 +1,27 @@
...
+This page details all of the Additional Checks in this instance of the Data Quality Tool. 
+
+The <label> Heading </label> is prefixed by text such as "X of your grants have..." in almost all cases, apart from when the check refers to the whole dataset (eg the multiple Funding Org check). The <label>Message</label> is displayed if the check is triggered. The <label>Method</label> describes how the check works under-the-hood. 
+</p>
+
+<h2> Checks </h2>
+{% for check in checks %}
+
+<p>
+<label>Heading</label> {{ check.heading }}</br>
```
edugomez: Templates in Cove do not have spaces between the curly braces and the variable name, so ```{{check.heading}}``` instead of ```{{ check.heading }}```. I personally prefer the latter, but not using spaces is more consistent  with the rest of the code.

cove_360/templates/cove_360/additional_checks.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% block content %}
+
+<div>
```
edugomez: In think this ```<div>``` is missing the closing ```</div>```.

cove_360/templates/cove_360/additional_checks.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.html' %}
```
edugomez: Given that this template was missing in a previous commit and the tests passed anyway, I wonder if we should have a simple test checking that this page is rendered correctly.

cove_360/templates/cove_360/additional_checks.html
Bjwebb: I've added it to our html validation check in  https://github.com/OpenDataServices/cove/pull/840/commits/b1d75acc9e8e5a3f6a5669b44d4b5cf91c8cedf2

Issue #320, opened 2016-04-01
Web API for Cove


Bjwebb:
As mentioned in https://github.com/open-contracting/standard/issues/319 there's some demand for a web API people can interact with programatically to get the same information Cove currently provides in a HTML page.

This could also be useful for us internally to run lots of files against Cove to check that different releases have the same behaviour (where expected).

This should be fairly straightforward to implement as a lot of the information already get serialized to JSON so that we can cache it.

See also [#207 - Calling Cove as a Python library (without the web interface)](https://github.com/OpenDataServices/cove/issues/207)

robredpath:
If/when we consider this, we should ensure that appropriate HTTP response codes are generated that reflect the result of the operations performed

robredpath:
@timgdavies @ekoner do you have any feel for how much demand there is for this?

ekoner:
@robredpath certainly demand from Budeshi and Spend Network (I believe). Internally from me as I often process numerous JSON files and either scrape or manually compile the results for report.

Tagging @SimKennedy for confirmation

robredpath:
@Bjwebb moving to Next Sprint - it seems like there's demand for it!

robredpath:
I'm going to try and get some more detail on this - the feeling from the dev team is that we want this to be a simple API but needs to be well-done.

Bjwebb:
We're going to start with a minimal implementation of an API in the next sprint - https://github.com/OpenDataServices/cove/issues/530 - which we can then share for feedback.

jpmckinney:
✅ From OCP's perspective, this issue can be closed / 'OCDS' tag can be removed. A web API is primarily for bulk processing. In that case, either the [jOCDS API](https://devgateway.github.io/jocds/jocds-web/#endpoints) can be used (which performs the structural checks defined by the schema and codelists), or lib-cove-ocds can be used as a library like in Kingfisher (which performs other calculations like additional fields, etc.). This is also the conclusion in #853. There is not sufficient demand for a web API around lib-cove-ocds, and I suspect it will take significant effort to make it performant enough for user expectations around web requests (the simplest optimization would be to make some calculations optional).

robredpath:
@jpmckinney Although this may not be a priority (which is fine with me!), I don't think that the idea of some method of automatic use of CoVE should be abandoned. 

[...]

PR #853, opened 2017-09-29
#320 JSON web output

patxiworks:
Hi @edugomez.

Sorry I had to create a fresh branch on this issue. The main reason is that the changes I'm proposing reflect what I would like to achieve: bypassing the django template and outputing the validation errors directly as json. For this, the validator needs to be fed with a separate url (http://standard.open-contracting.org/validator/?raw=true&source_url=XXXX) and redirected to another (http://standard.open-contracting.org/validator/raw/xxxx...) which displays the errors as json.

Unfortunately, I couldn't implement this with cove_ocds/lib/api.py as you suggested. But I succeeded in eliminating previously duplicated code. Changes are now fewer but now exclusively done in core files (just 4 of them).

Looking forward to feedback.


cove_ocds/views.py
```diff
@@ -15,6 +16,7 @@
 from cove.lib.exceptions import CoveInputDataError, cove_web_input_error
 from cove.views import explore_data_context
 
+from cove_ocds.webservice import views
```
Bjwebb: This module doesn't exist.

patxiworks: The module has been taken out. It was a remnant that slipped through from the previous version. Please how can I check up the Travis CI build errors?

edugomez: >  Please how can I check up the Travis CI build errors?
You can check Travis build errors by clicking on the red X in the latest commit at which the build failed. You will be redirected to travis-ci.org/OpenDataServices..etc, where you will find two builds (Python 3.4 and 3.5). Clicking on the builds will show you the logs with the errors.

patxiworks: Thanks @edugomez. I've had a look but perhaps a summary of the errors would be more useful at this point.

edugomez: @patxiworks you can ignore this error ```test_url_input[tenders_releases_1_release_with_extensions_new_layout.json-expected_text2-not_expected_text2-True]```, it is not due to your code (it will be fixed when PR #852 gets merged -- this commit specifically https://github.com/OpenDataServices/cove/pull/852/commits/5dc875ca54fe4fd50abe36a1c8c021872a58dcc0)

edugomez:
@patxiworks if you pull latest master to your fork, merge into the PR branch and update the PR Travis _may_ stop complaining

cove_ocds/views.py
```diff
@@ -159,5 +159,14 @@ def explore_ocds(request, pk):
...
-
-    return render(request, template, context)
+    
+    # Check the path to know whether a 'raw' request has been made
+    request_path = request.path
+    if 'raw' in request_path.split('/'):
+        response = {}
+        response['validation_errors'] = context['validation_errors']
+        response['validation_errors_count'] = context['validation_errors_count']
+        # Output as raw json
```
edugomez: I think we should probably output the whole output of ```ocds_json_output``` at _cove/cove_ocds/lib/api.py_. Given that we don't have real api endpoints, we could give a response with all we've got and let the requester take from it what they need.

Please note that ```ocds_json_output``` validation errors are output a bit differently to the one in the view ```context```, so this ```context['validation_errors']```:

```
[('["date-time", "Date is not in the correct format", "publishedDate"]', [{'value': '2014-11-07', 'path': 'publishedDate'}]), ('["date-time", "Date is not in the correct format", "releases/date"]', [{'value': '2014-03-25', 'path': 'releases/0/date'}, {'value': '2014-04-04', 'path': 'releases/1/date'}]), ('["date-time", "Date is not in the correct format", "releases/tender/tenderPeriod/endDate"]', [{'value': '2014-04-15 14:00 Eastern Daylight Time (EDT)', 'path': 'releases/0/tender/tenderPeriod/endDate'}, {'value': '2014-04-22 14:00 Central Daylight Time (CDT)', 'path': 'releases/1/tender/tenderPeriod/endDate'}]), ('["required", "\'id\' is missing but required", "releases"]', [{'path': 'releases/0'}]), ('["uri", "Invalid \'uri\' found", "publisher/uri"]', [{'value': 'buyandsell.gc.ca', 'path': 'publisher/uri'}])]
```
will look like this in the cli output version of it:
```
...
  "validation_errors": [
    {
      "value": "2014-11-07",
      "type": "date-time",
      "path": "publishedDate",
      "description": "Date is not in the correct format",
      "field": "publishedDate"
    },
    {
      "value": "2014-03-25",
      "type": "date-time",
      "path": "releases/0/date",
      "description": "Date is not in the correct format",
      "field": "releases/date"
    },
    {
      "value": "2014-04-04",
      "type": "date-time",
      "path": "releases/1/date",
      "description": "Date is not in the correct format",
      "field": "releases/date"
    },
    {
      "value": "2014-04-15 14:00 Eastern Daylight Time (EDT)",
      "type": "date-time",
      "path": "releases/0/tender/tenderPeriod/endDate",
      "description": "Date is not in the correct format",
      "field": "releases/tender/tenderPeriod/endDate"
    },
    {
      "value": "2014-04-22 14:00 Central Daylight Time (CDT)",
      "type": "date-time",
      "path": "releases/1/tender/tenderPeriod/endDate",
      "description": "Date is not in the correct format",
      "field": "releases/tender/tenderPeriod/endDate"
    },
    {
      "value": "",
      "type": "required",
      "path": "releases/0",
      "description": "'id' is missing but required",
      "field": "releases"
    },
    {
      "value": "buyandsell.gc.ca",
      "type": "uri",
      "path": "publisher/uri",
      "description": "Invalid 'uri' found",
      "field": "publisher/uri"
    }
  ]
...
```
And it won't include any aggregations, that will be up to the consumer of the output to do (it won't include a ``` 'validation_errors_count'``` either)

Hope this makes sense.

cove/input/views.py
```diff
@@ -82,6 +82,10 @@ def data_input(request, form_classes=default_form_classes, text_file_name='test.
                     })
             elif form_name == 'text_form':
                 data.original_file.save(text_file_name, ContentFile(form['paste'].value()))
-            return redirect(data.get_absolute_url())
+            
+            if 'raw' in request_data:
```
edugomez: How does ```'raw'``` make it to ```request_data```? Are you sending a query string in the URL to the GET request when asking for the json response? If so, how? Manually, adding it to the URL for the existing data URL?

Also, note that you are not adding parameters to the upload forms, which means it won't be a ```'raw'``` in any of the POST requests.

cove_ocds/urls.py
```diff
@@ -10,7 +10,10 @@
 
 
 # Serve the OCDS validator at /validator/
-urlpatterns_core += [url(r'^data/(.+)$', cove_ocds.views.explore_ocds, name='explore')]
+urlpatterns_core += [
+    url(r'^data/(.+)$', cove_ocds.views.explore_ocds, name='explore'),
+    url(r'^raw/(.+)$', cove_ocds.views.explore_ocds, name='explore_raw')
```
edugomez: Hmmm, two URLs mapping  to the same view. I'm not sure, it may be user cases for this, but I don't see it here. I understand that you are doing this so in the view you check for the URL with ```raw``` in it, but I don't think that is the best way to do so. If you want a view behaving differently for different requests, that is normally done using GET and POST parameters  https://docs.djangoproject.com/en/1.11/ref/request-response/

cove_ocds/views.py
```diff
@@ -159,5 +159,14 @@ def explore_ocds(request, pk):
             context['releases'] = json_data['releases']
         else:
             context['releases'] = []
-
-    return render(request, template, context)
+    
+    # Check the path to know whether a 'raw' request has been made
+    request_path = request.path
+    if 'raw' in request_path.split('/'):
```
edugomez: See my comment above, there are better ways to do this.

cove/input/models.py
```diff
@@ -44,6 +44,9 @@ class SuppliedData(models.Model):
 
     def get_absolute_url(self):
         return reverse('explore', args=(self.pk,), current_app=self.current_app)
+    
+    def get_raw_url(self):
```
edugomez: See my comments below, you may not need this. Instead you can pass extra arguments to ```get_absolute_url```

edugomez:
@patxiworks be aware that I have merged in changes from master, so please pull again before any additional work.

cove_ocds/urls.py
patxiworks: @edugomez, this makes sense. Should have treated this in the same way as 'source_url' in GET (in fact, there's already a raw parameter sent with the url).

cove/input/models.py
patxiworks: Yes, this is true.

cove_ocds/views.py
patxiworks: Yes, this makes sense. Only issue has to do with accessing `ocds_json_output` from `explore_ocds`. I think I already tried to do that but couldn't retrieve the validation errors produced by `ocds_json_output`. That's why I resorted to pulling them from `context`.

Any suggestions?

edugomez:
@patxiworks, my doubt is how are you thinking to request the data (sorry if you have mentioned this before and I missed it):
1 - If you intend to send a request using e.g. Python's ```requests``` library (or PHP's ```cURL```, etc...), then it should be easy, as you just need to add a parameter (say ```'raw':True```) to the POST (when you upload data to validate) or GET (when you want to retrieved previously uploaded data, for which you need to know the URL) requests. Then you have to look for the parameter in the ```request.POST``` or ```request.GET```dictionary-like objects passed to the view.

2 - If you want to get json in the browser, that it's a bit more complex, as the user needs to specify in the form the output they want, or there' should be a button or something in the explore page that would return the json data, or both.

If you are not so familiar with Django and do not have the time to be, I could have a go at option 1 (but I don't want to step on your toes, so please go ahead if convenient). We could have a chat in hangouts if useful too.

cove_ocds/views.py
edugomez: I think this would work:
```
from cove_ocds.lib.api import context_api_transform
... 
    response = context_api_transform(context)
...
```
Validation errors should be ```response['validation_errors']```

patxiworks:
@edugomez, sorry for my late reply.

I think I'd prefer to go with option 2 because it is more generic for applications that would like to access the json regardless of the language used. Here's the use case I'm interested in:

The user points the browser to a URL similar to this: `http://standard.open-contracting.org/validator/?raw=true&source_url=XXXX` (where XXXX is the URL of the OCDS data to be validated). The validator then displays the validation errors as JSON in the browser. In this case, any consuming application can just make a request (for example, using `cURL` for PHP or something similar if using another language) through the URL above and retrieve the validation errors from the browser output. This way, it would also be possible to just enter the appropriate URL directly in the browser and copy out the raw validation errors.

The code I'm proposing already works like this. But to make it better, I need to implement the changes you've suggested (i.e to use `request.GET` and `context_api_transform(context)`).

I hope this makes things clearer.

edugomez:
@patxiworks Hi, how is it going? 

Quick update on this issue:

I think a very quick solution for the time being would be to follow the approach for CoVE IATI by @kindly in this PR #907 (for POST requests) and then do a request programmatically like this (for Python):
``` 
import requests
r = requests.post('http://host/validator/api_test', files={'file': open(~/example.json', ' rb')}, data={'name':'any.json'})
r.json()
```
As for the GET request in the browser, _/validator_  is already accepting a query string param ```source_url```, so it needs to be passed the extra ```raw``` parameter you proposed to return the json data instead. This would require some modification to the way of doing here for IATI https://github.com/OpenDataServices/cove/blob/905-post-json/cove_iati/views.py#L94 (as it is accepting POST requests only)

As I mentioned before, I wouldn't implement POST requests in the browser to get JSON back as it requires more work.

edugomez:
@patxiworks I'm considering closing this PR as I'm afraid it may slowly be wilting here. Is that OK?

patxiworks:
Hi @edugomez. Really sorry for the slow response from me.

Indeed, the project was wilting at my end but I just got a reason to give
it another go. Please could we leave it open so that I could quickly
implement your latest suggestions?

Thanks!

edugomez:
@patxiworks sure, as long as you are still interested it can stay open!

Issue #859, opened 2017-10-04
Update requirements


Bjwebb:
Get the latest versions of the python libraries we rely on, in order to get any bug fixes that may have been made.

robredpath:
Duplicates #603

PR #865, opened 2017-10-11
[#859] Update requirements

Bjwebb:
#859


Bjwebb:
Loss of coverage is substantial. I think we need to review that before merging.

PR #867, opened 2017-10-12
Replace Organisation Registration Agency with Org-ids lists

edugomez:


edugomez:
@Bjwebb can you approve this now if the changes are OK?

Issue #855, opened 2017-10-03
XML -> spreadsheet conversion


Bjwebb:
https://github.com/OpenDataServices/flatten-tool/issues/151

Bjwebb:
I have the basic functionality working, and deployed to a dev copy of cove: http://855-xml-to-spreadsheet.dev.cove.opendataservices.coop/

There's a list of things I know I need to check/test in the flatten-tool issue: https://github.com/OpenDataServices/flatten-tool/issues/151#issuecomment-333869766

PR #881, opened 2017-10-18
IATI XML to Spreadsheet conversion

Bjwebb:
#855


cove/lib/converters.py
```diff
@@ -112,7 +112,7 @@ def convert_spreadsheet(upload_dir, upload_url, file_name, file_type, schema_url
     return context
 
 
-def convert_json(upload_dir, upload_url, file_name, schema_url, replace=False, request=None, flatten=False, cache=True):
+def convert_json(upload_dir, upload_url, file_name, schema_url=None, replace=False, request=None, flatten=False, cache=True, xml=False):
```
edugomez: No big deal, but is this function name kind of a misnomer now?

PR #888, opened 2017-11-07
agreed text updates to additional checks

BobHarper1:


cove_360/lib/threesixtygiving.py
```diff
@@ -439,8 +439,8 @@ class IncompleteRecipientOrg(AdditionalTest):
     """Checks if any grants lack one of either Recipient Org:Postal Code or both of Recipient Org:Location:Geographic Code and Recipient Org:Location:Geographic Code Type"""
 
     check_text = {
-        "heading": "incomplete recipient organisation information",
-        "message": "Your data is missing either Recipient Org: Postal Code or Recipient Org: Location:Geographic Code combined with Recipient Org: Location: Geographic Code Type. Knowing the geographic location of recipient organisations allows users of your data to understand your data and combine it with other data sets to form a broader picture of grant-making."
+        "heading": "not have recipient organisation location information",
```
robredpath: @BobHarper1 you'll need to change the verb in order for this to make sense. See https://github.com/OpenDataServices/cove/blob/master/cove_360/lib/threesixtygiving.py#L434 for an example .

cove_360/lib/threesixtygiving.py
```diff
@@ -505,7 +505,7 @@ class LooksLikeEmail(AdditionalTest):
 
     check_text = {
         "heading": "text that looks like an email address",
-        "message": "This may indicate that the data contains personal data, use and distribution of which is restricted by the Data Protection Act. You should ensure that any personal data is included with the knowledge and consent of the person to whom it refers."
+        "message": "An @ symbol has been found in your data. This may indicate that the data contains an email address, which can constitute personal data. The use and distribution of personal data  is restricted by the Data Protection Act. You should ensure that any personal data is only included with the knowledge and consent of the person to whom it refers."
```
robredpath: This isn't quite true, because "pets@home" wouldn't trigger this check, nor would a Twitter handle. It looks for a string with some text, then an @, then some more text that must include at least one dot. If this check has triggered, it almost certainly is because of an email address.

cove_360/lib/threesixtygiving.py
```diff
@@ -288,10 +288,10 @@ def process(self, grant, path_prefix):
 
 
 class RecipientOrgUnrecognisedPrefix(AdditionalTest):
-    """Check if any grants have RecipientOrg IDs that use a prefix that isn't an Org-id registered prefix"""
+    """Check if any grants have RecipientOrg IDs that use a prefix that isn't on the Org ID prefix codelist"""
 
     check_text = {
-        "heading": "a Recipient Org:Identifier that doesn’t draw from an external identification body",
+        "heading": "a Recipient Org:Identifier does not draw from a recognised register.",
```
robredpath: I don't think we can remove the 'that' from this sentence

robredpath:
Thanks for starting this off, @BobHarper1 ! Looks good on the whole, just a couple of things I've commented on directly, and you'll need to update the tests with the new wording - you'll find them in https://github.com/OpenDataServices/cove/blob/6944e9ed04fad71512a78c717b9345f370fe8381/cove_360/tests.py

BobHarper1:
Thanks @robredpath - I'll do those changes and fix the tests.

cove_360/lib/threesixtygiving.py
BobHarper1: verb added "do"

BobHarper1:
Wee note, I can't run CoVE and do tests locally e.g. trying to run migrate I get a traceback ending

```python
....
  File "/mnt/c/Users/bobha/Documents/opendataservices/cove/cove_360/views.py", line 82
    context["checks"] = [{** check.check_text, 'desc': check.__doc__} for check in TEST_CLASSES]
                           ^
SyntaxError: invalid syntax
```

Bjwebb:
@BobHarper1 I think that's new syntax that's introduced in Python 3.5. I'm guessing that you're using an older version?

BobHarper1:
Thanks @bjwebb, yes 3.4

On 8 Nov 2017 4:56 p.m., "Ben Webb" <notifications@github.com> wrote:

@BobHarper1 <https://github.com/bobharper1> I think that's new syntax
that's introduced in Python 3.5. I'm guessing that you're using an older
version?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/OpenDataServices/cove/pull/888#issuecomment-342882030>,
or mute the thread

robredpath:
@BobHarper1 CoVE requires Python 3.5 at a minimum - I'm sure @Bjwebb and @kindly have installed it on Ubuntu before. I've only ever installed it on CentOS so can't really advise!

Bjwebb:
As an aside, `branch-360-additional-text-update`, is a bit of an odd name for this branch. My branches start with `branch-360-` because they're about deploying a specific branch of the 360 standard, whereas this isn't. No need to change it this time though.

BobHarper1:
Ah, ok I'll try an remember for the future

Bjwebb:
This is now deployed to the dev site http://master.cove-360-dev.default.threesixtygiving.uk0.bigv.io/

PR #889, opened 2017-11-08
Refactor templates

robredpath:


robredpath:
This PR should result in no change to the output. I've run a validation with all three CoVEs and eyeballed the output in each case, and I think we're good. Additional eyes welcome.

robredpath:
Fixed @Bjwebb

Bjwebb:
More info box appear on the terms and conditions page, and the 360Giving common errors page, where it didn't before.

robredpath:
@Bjwebb @edugomez could you take a look at this? I've checked over the output but additional eyes are always useful. 

The 'how to use' box disappears from IATI CoVE, but that's intentional - until we get some content to actually put in it (which is the job that started all of this off!)

PR #893, opened 2017-11-14
 Prevent ruleset errors from causing a 500 internal server error

Bjwebb:
Original sentry issue: https://sentry.io/open-data-services/cove-dev/issues/392791032/


Bjwebb:
I think what's missing is an indication in the web interface that ruleset checks have failed, rather than them all having passed.

edugomez:
> I think what's missing is an indication in the web interface that ruleset checks have failed, rather than them all having passed.

I don't think the web interface will say that. Instead, it will show a single message ('There was a problem running the ruleset checks.') for a single error (with a cross not a tick in the section heading).

Still, not ideal, we should change for something else.

cove_iati/lib/iati.py
```diff
@@ -44,14 +45,25 @@ def common_checks_context_iati(context, upload_dir, data_file, file_type, api=Fa
         schema = lxml.etree.XMLSchema(schema_tree)
         schema.validate(tree)
         lxml_errors = lxml_errors_generator(schema.error_log)
-        ruleset_errors = get_iati_ruleset_errors(tree, os.path.join(upload_dir, 'ruleset'))
+        ruleset_errors = get_iati_ruleset_errors(
+            tree,
+            os.path.join(upload_dir, 'ruleset'),
+            ignore_errors=not api,
+            return_on_error=[{'message': 'There was a problem running the ruleset checks.'}],
```
edugomez: We need to pass some info to the template so that we render this differently from common ruleset errors, as  this message says that it wasn't possible to run the ruleset check errors.

Also, the message is a bit obscure, perhaps we need to give more context for the error.

edugomez: @Bjwebb I was thinking, I'm not sure we should be doing this. The problem in this specific issue is due to:

 ```attr_sum = sum(Decimal(x.attrib.get(attribute, 0)) for x in elements)``` (notice that ```x.attrib = ""``` does not default to ```0```, which is the case when having ```percentage=""``` in the data)

Probably  we should make sure we pass only digits to ```Decimal``` and fix the problem there. Next time we have an exception in the code, I think it would be better to get the 500 (which the user can report to us if they want) instead of a 200 and a cryptic message. That way we will be made aware of the error so we can fix rather than ignoring it.

What do you think?

edugomez: After chat with @Bjwebb, the best strategy seems to swallow exceptions in ruleset checks if the data does not pass validation (but throw 500 if ruleset checks break with valid data).This is the same behaviour in other places in Cove (Cove OCDS https://github.com/OpenDataServices/cove/blob/master/cove_ocds/lib/ocds.py#L352)

edugomez:
@Bjwebb this PR is almost ready, I just need to add some tests (slightly tricky) . Do you think you could deploy this branch on a dev server and ping @stevieflow to have a go at it?

Bjwebb:
@edugomez Deployed now to http://ruleset-no-500.dev.cove.opendataservices.coop/. Pinging @stevieflow  to test.

edugomez:
@Bjwebb Amazing, thanks!

edugomez:
@Bjwebb can you take a look at this PR whenever you find the time?

stevieflow:
OK thanks.  This seems to covert.  I got a load of the Duplicate Identifier messages back - but the XML was valid...

Issue #905, opened 2017-11-30
Create POST endpoint for IATI CoVE to get JSON output


rory09:
The aim is to create an endpoint where you can POST a file and get back the same JSON results as the CLI.
For now the endpoint is at ```/api_test```
So you need to post a file with parameter name "file" and a string parameter "name" indicating the name of the file with the extension i.e myfile.xlsx.  It should support xml, csv and xlsx.

PR #907, opened 2017-12-06
[#905] Add post API endpoint for IATI

kindly:


cove_iati/tests.py
```diff
@@ -518,3 +518,25 @@ def test_common_checks_context_iati_ruleset():
     upload_dir = os.path.join('media', str(uuid.uuid4()))
     context = iati.common_checks_context_iati({}, upload_dir, file_path, 'xml')
     assert len(context['ruleset_errors']) == 17
+
+
+@pytest.mark.django_db
```
edugomez: I think pytest does not need to use database access here

cove_iati/views.py
```diff
@@ -76,3 +89,19 @@ def explore_iati(request, pk):
...
+@csrf_exempt
+def api_test(request):
+    form = UploadApi(request.POST, request.FILES)
+    if form.is_valid():
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            file_path = os.path.join(tmpdirname, form.cleaned_data['name'])
+            with open(file_path, 'wb+') as destination:
+                for chunk in request.FILES['file'].chunks():
+                    destination.write(chunk)
+            result = iati_json_output(tmpdirname, file_path, form.cleaned_data['openag'])
```
edugomez: I just noticed it doesn't give the option of doing org-ids checks. Can we add it to the form and pass it to ```iati_json_otuput```?

PR #912, opened 2017-12-11
Folder option in ocds-cli tool

yolile:
Adds the option to evaluate a folder of json files and save the validation
results into a result.csv file.


jpmckinney:
Instead of new `--file` and `--folder` options, can the code just test whether the input `file` is a folder (`os.path.isdir()`)? That might avoid the test failures as well.

yolile:
Done!

jpmckinney:
Not sure why coveralls is upset, but tests pass.

README.rst
```diff
@@ -232,12 +232,14 @@ CoVE for OCDS and IATI can be run from the command line. To get a JSON file with
...
 
-    ./ocds-cli --options file-name
+    ./ocds-cli --options
+
 
-``file-name`` can be a JSON or an Excel file.
 
 Options:
 
+``--file-name -file`` can be a JSON or an Excel file or folder.
```
jpmckinney: I think the documentation change here is incorrect. `file-name` is a mandatory argument (like before the changes). It's not an option.

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +31,28 @@ def handle(self, file, *args, **options):
             ))
 
         try:
-            result = ocds_json_output(self.output_dir, file, schema_version, convert)
+            if os.path.isdir(file):
```
jpmckinney: The behavior between the file and directory are not the same. A file produces a full `results.json` file, but a directory produces a CSV with only the validation errors.

I would recommend having a directory produce a JSON Lines file (one line of JSON for each file). An external script can read that the JSON Lines file to produce a CSV if that's what's desired.

jpmckinney (changes requested):

yolile:
I made the suggested changes

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +30,23 @@ def handle(self, file, *args, **options):
...
+                directory = os.fsencode(file)
+                outfile = open(os.path.join(self.output_dir, "all_results.txt"), 'w+')
+                for f in os.listdir(directory):
+                    filename = os.fsdecode(f)
+                    if filename.endswith(".json"):
+                        result = ocds_json_output(self.output_dir, file + filename, schema_version, convert)
```
jpmckinney: Instead of `+`, use `os.path.join`

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +30,23 @@ def handle(self, file, *args, **options):
             ))
 
         try:
-            result = ocds_json_output(self.output_dir, file, schema_version, convert)
+            if os.path.isdir(file):
+                directory = os.fsencode(file)
```
jpmckinney: What is the reason for `fsencode`?

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -1,7 +1,6 @@
 import json
 import os
 import sys
-
```
jpmckinney: Restore the newline.

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +30,23 @@ def handle(self, file, *args, **options):
...
+                directory = os.fsencode(file)
+                outfile = open(os.path.join(self.output_dir, "all_results.txt"), 'w+')
+                for f in os.listdir(directory):
+                    filename = os.fsdecode(f)
+                    if filename.endswith(".json"):
+                        result = ocds_json_output(self.output_dir, file + filename, schema_version, convert)
+                        result.update({'file': filename})
+                        json.dump(result, outfile, cls=SetEncoder)
+                        outfile.write("\n")
+                    else:
```
jpmckinney: We don't need this `else` branch.

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +30,23 @@ def handle(self, file, *args, **options):
             ))
 
         try:
-            result = ocds_json_output(self.output_dir, file, schema_version, convert)
+            if os.path.isdir(file):
+                directory = os.fsencode(file)
+                outfile = open(os.path.join(self.output_dir, "all_results.txt"), 'w+')
```
jpmckinney: Use `with open` syntax, otherwise if there's an exception, `outfile.close()` will not be reached.

yolile:
done

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +31,21 @@ def handle(self, file, *args, **options):
             ))
 
         try:
-            result = ocds_json_output(self.output_dir, file, schema_version, convert)
+            if os.path.isdir(file):
+                with open(os.path.join(self.output_dir, "all_results.txt"), 'w+') as outfile:
+                    for f in os.listdir(file):
+                        filename = os.fsdecode(f)
+                        if filename.endswith(".json"):
```
edugomez: No big deal, but the code in this repo normally uses single quotations marks (although TBH that is not 100% consistent, with myself one of the persons to blame)

edugomez:
@yolile any chance you could add a test for this new functionality? You can find some examples here https://github.com/OpenDataServices/cove/blob/master/cove_ocds/tests.py

kindly:
@yolile this is a great start and I think it is something we should do.

I am not convinced however this is the correct way we should accept multiple files.
Instead of selecting either a folder or a file we should accept a list of files i.e

```
ocds-cli file1.json file2.json file3.json
```
This way you can do things in the command line like:
```
ocds-cli directory/*.json
```
and your shell will expand out the list of files for you.  This will give the user much more flexibility as to what files to include and is a more standard way of selecting multiple files. You can do more complicated selctions too.

```
ocds-cli directory/a*.json directory/b*.json
```
i.e all files in that directory that start with a or b

The main reason I have issues with just selecting a directory is that if that directory contains any files that are not OCDS files, the command will fail.  This is a common problem i.e you could have a .git folder in there or a bit of documentation and you want or your text editor puts a lock file in there.  We could filter out just .json or .xlsx or .csv files but I think allowing the user to select the exact files they want is better.

I understand doing it the above way is a lot more work and means we have to modify how this works for the other standards cove supports but I think it is worth it.

On top of this we need to give the user some nicer feedback as to what file failed when they are selecting multiple files by maybe printing out each file as we are processing them or catching the error and printing out the filename then.

kindly:
@jpmckinney @yolile 

I am not convinced that the output should be a JSON Lines file.  I would be much happier with just a list of objects i.e
```
[
 {result 1},
 {result 2},
 ...
]
```
So the file is still just a normal JSON file.  This makes it much easier to parse using any language without a special JSON Lines library. Also it means the results can be more human readable i.e we do not have to fit each result in a single row. 
There are streaming json libraries i.e ijson for python, that can stream lists if needed so I do not see the benefit of JSON Lines for this use case.

I think we should just have one output file called results.json and have its contents a JSON list if you supply multiple files and an object if you supply one.

edugomez:
> I think we should just have one output file called results.json and have its contents a JSON list if you supply multiple files and an object if you supply one.

:+1: This behaviour sounds much more consistent with what the CLI currently does

jpmckinney:
@kindly @edugomez JSON Lines just requires splitting on newlines and then reading the line with any JSON library. It doesn't require a specialized library.

Having an array in one case and an object in the other case is not desirable, because now you need different logic for different result files. The options for consistency are either: (1) have an array in all cases, i.e. always treat the input as if it is potentially globbing files; (2) JSON Lines (I don't know why we should force the use of a specialized library like ijson to do streaming).

The value of streaming in general, is that, ideally, I should be able to write to STDOUT instead of any file, and pipe that output to another script that starts reading the results and doing some work on it.

Also, JSON Lines makes it easy to skip/cut validation reports using simple tools like `head` and `tail`. There are many other things I'd like to do with this output for which an array is inappropriate.

The use case of a human reading a giant JSON file with 100,000 validation reports (which they in all likelihood cannot open within a reasonable time with their text editor) is not a real.

If you want, we can implement a `--stream` option which outputs JSON Lines, and that otherwise writes a human-readable array (in all cases, not just when multiple files are passed).

kindly:
@jpmckinney happy to have a --stream option that would print to stdout in JSON lines. It makes perfect sense in that case. It would be up to the user then to pipe it to where they want it to be, which could be a file. If it is a single file input then that would work too and just produce a single JSON object.

We could have a jsonlines option also which would make the file in this pull request but would probably call it results.jsonl as they should be separate files. However, this is less useful than stream option.

I think a more common use case (even though its not yours) would be validating over a handful of files and then you want human readability. I am considering having an array in all cases of the file but wary of backwards incompatibility doing that and is why I suggested the above. I personally do not think its a big burden to anyone using the cli and would prefer backwards compatibility.

jpmckinney:
Sounds good.

I think the number of people using the CLI (outside OCP and helpdesks) is close to zero, as it is not well advertised, so I think it's okay to be backwards incompatible. You can put a notice in the readme, if you like, that the output is unstable because the feature is alpha (or however you want to put it). At this stage I think we should prefer consistency and simplicity over backwards compatibility.

kindly:
After looking at this more closely I remembered why initially we chose to only support a single file as input.
When converting, from xlsx or csv we store the converted file for people to use in the created directory.  If we accept multiple files this will be difficult to do as we have to create some kind of namespacing in the folder to make sure the converted files do not name clash (at the moment we just call the file unflattened.json).  We also (optionally) copy over the file you supplied over to the folder to you have the file with your results and this will need namespacing too.
We could say that if you supply multiple files then you do not get this functionality but this is very inconsistent behaviour also.  Also, on top of this other standards use this feature more and have more use of the CLI so breaking backwards compatibility for them is more problematic.

Nonetheless I think the ```--stream``` option is useful. So the behaviour I think would should do for now is.

- The command line accepts multiple files.
- The ```--stream``` option will exist and print out jsonlines to stout and not produce any output folder.
- If you select multiple files without the ```--stream``` option then you will get an error explaining that you need to use it.
- you can still use the ```--stream``` option with a single file and get a single line jsonlines output and no folder structure.

I think this covers the need expressed here and can be implemented by all the standards cove supports with no issues.
We could discuss a way to support multiple files when outputting a folder but I think that is a much bigger issue.

yolile:
everybody agree with what @kindly  proposes? so I can make the changes for it
@jpmckinney @edugomez ?

edugomez:
> Nonetheless I think the --stream option is useful. So the behaviour I think would should do for now is.
> 
> - The command line accepts multiple files.
> - The --stream option will exist and print out jsonlines to stout and not produce any output folder.
> - If you select multiple files without the --stream option then you will get an error explaining that you need to use it.
> - you can still use the --stream option with a single file and get a single line jsonlines output and no folder structure.

@yolile All that sounds good to me. The only thing I would add is a reminder to document that behaviour in the  [README.str file](https://github.com/OpenDataServices/cove/blob/master/README.rst#command-line-interface) as clearly as possible, as the output with the ```--stream``` option will be very different to the one without.

jpmckinney:
Fine for this to go ahead.

For the non-stream multiple file issues, why would the output differ from running each file individually through the command? The command already namespaces folders, by using the basename of the file to process. I'm not sure why we would just do that in a first instance.

jpmckinney:
Also, can we configure coveralls to have a small range of acceptable differences in coverage?

edugomez:
> The command already namespaces folders

That's right.

> I think the issues only arise if we want a single output folder for all input files, which hasn't been expressed as a requirement

The one problem I see with this is that passing a large number of files to the command would create a large number of new directories, which can be a bit annoying. I think it would be neater if the command created a parent directory for all of them. I wonder if instead of this: 

_If you select multiple files without the --stream option then you will get an error explaining that **you need to use it**._

we should add a new option  ``--parent-dir``:

_If you select multiple files without the --stream option then you will get an error explaining that **you need to choose a name for the parent directory using --parent-dir**._ (in this case ```--output-dir```, if included, would be ignored, perhaps with a warning)

which would eliminate the asymmetric behaviour in @kindly 's proposal.

edugomez:
I'm not sure what's the deal with coveralls in this PR saying same coverage remains and still failing, (@Bjwebb knows better than me about coveralls), but in any case we should add news tests for all the planned changes to the command line (which so far is reasonably well tested).

kindly:
@jpmckinney if you run each file individually through the command the user gets to choose the output folder.  The command defaults the folder name to the name of the file  (minus extension) but there is an option to specify a different folder if you want. 
On selecting multiple files we have to assume all file names are unique and that may not be the case. So we have to come up with a naming scheme if there is a clash or throw an error if that happens. Both of these options seem bad to me or need some discussion at least.

My thinking is for this case for just a few files not streamed would be very easy for a user to do a little bash script for and then they would have full control over what folders are created.

@edugomez @jpmckinney Concerning coveralls. I am happy for it to fail as it does not block us merging and acts more like a warning. For this case, it can only not be merged as its pending review. I think it is right that coveralls is failing for this case as we have more code not tested but I am not sure why it is not showing that.

jpmckinney:
Sure, let's open a new issue for non-stream output and focus on the stream output in this issue – around which there's agreement.

yolile:
I added the --stream option and the multiple files / regex expression option, updated the README and made the tests. @jpmckinney @edugomez @kindly

README.rst
```diff
@@ -232,9 +232,10 @@ CoVE for OCDS and IATI can be run from the command line. To get a JSON file with
 
 .. code:: bash
 
-    ./ocds-cli --options file-name
+    ./ocds-cli --options files
 
-``file-name`` can be a JSON or an Excel file.
+``files`` can be a list of JSON or Excel files. Regular expressions are supported. The use of multiple files must be
```
edugomez: I think this is misleading. The command accepts a single file or a directory containing a number of files, so I would leave it as ```file-name``` adding that multiple files are also accepted (```e.g, dir-name/*.json```) passing the option ```--stream```

cove/management/commands/base_command.py
```diff
@@ -19,28 +20,37 @@ def __init__(self, *args, **kwargs):
...
+        parser.add_argument('files', help='Files to be processed by Cove', nargs='+')
         parser.add_argument('--output-dir', '-o', default='', help='Directory where the output is created, defaults to the name of the file')
         parser.add_argument('--delete', '-d', action='store_true', help='Delete existing directory if it exits')
         parser.add_argument('--exclude-file', '-e', action='store_true', help='Do not include the file in the output directory')
+        parser.add_argument('--stream', '-st', action='store_true', help='Do not include the file in the output directory')
+
+    def handle(self, files, *args, **options):
+
+        stream = options.get('stream')
+        if (len(files) > 1 or len(glob.glob(files[0])) > 1 or os.path.isdir(glob.glob(files[0])[0])) and not stream:
```
edugomez: ```if len(files) > 1 and not stream:``` should be enough.

Also, ```len(glob.glob(files[0]) > 1``` and ```os.path.isdir(glob.glob(files[0])[0])``` don't make much sense. @yolile I can see that perhaps you misunderstood what ```glob.glob``` does?

cove/management/commands/base_command.py
```diff
@@ -19,28 +19,37 @@ def __init__(self, *args, **kwargs):
...
         parser.add_argument('--output-dir', '-o', default='', help='Directory where the output is created, defaults to the name of the file')
         parser.add_argument('--delete', '-d', action='store_true', help='Delete existing directory if it exits')
         parser.add_argument('--exclude-file', '-e', action='store_true', help='Do not include the file in the output directory')
+        parser.add_argument('--stream', '-t', action='store_true', help='The output would be streamed to stdout')
+
+    def handle(self, files, *args, **options):
+
+        stream = options.get('stream')
+        if (len(files) > 1 or os.path.isdir(files[0])) and not stream:
+            self.stdout.write('The multiple file option must be used with the --stream option')
```
kindly: I think we need for streaming to print every error out to stderr not stdout.  This is because the user will expect the stdout to be in JSON lines format.  If we print out anything non JSON to stdout then it will cause the user an issue.

cove_iati/management/commands/iati_cli.py
```diff
@@ -17,19 +16,28 @@ def add_arguments(self, parser):
...
         try:
-            result = iati_json_output(self.output_dir, file, openag=openag, orgids=orgids)
+            if stream:
+                for file in files:
+                    if os.path.isdir(file):
+                        self.stdout.write('Skipping %s directory ' % str(file))
+                    else:
+                        result = iati_json_output(self.output_dir, file, openag=openag,
+                                                  orgids=orgids)
+                        self.stdout.write(str(result))
```
kindly: The result should be JSON dumped not just print the string representation in python as they are different i.e the str representation can use single quotes for strings.

cove_iati/management/commands/iati_cli.py
```diff
@@ -17,19 +16,28 @@ def add_arguments(self, parser):
...
+        stream = options.get('stream')
         openag = options.get('openag')
         orgids = options.get('orgids')
 
         try:
-            result = iati_json_output(self.output_dir, file, openag=openag, orgids=orgids)
+            if stream:
+                for file in files:
+                    if os.path.isdir(file):
+                        self.stdout.write('Skipping %s directory ' % str(file))
```
kindly: should be stderr also

cove_iati/management/commands/iati_cli.py
```diff
@@ -17,19 +16,28 @@ def add_arguments(self, parser):
...
+                    else:
+                        result = iati_json_output(self.output_dir, file, openag=openag,
+                                                  orgids=orgids)
```
kindly, line 32: I think there should not be any need for an output directory when using streaming.
I think a temporary directory should be made instead, that is automatically deleted afterwards.  We do this for the IATI api here.
https://github.com/OpenDataServices/cove/blob/master/cove_iati/views.py#L100

cove_iati/management/commands/iati_cli.py
```diff
@@ -17,19 +16,28 @@ def add_arguments(self, parser):
...
+                        self.stdout.write(str(result))
+            else:
+                result = iati_json_output(self.output_dir, files[0], openag=openag, orgids=orgids)
+                with open(os.path.join(self.output_dir, "results.json"), 'w+') as result_file:
+                    json.dump(result, result_file, indent=2, cls=SetEncoder)
         except APIException as e:
             self.stdout.write(str(e))
```
kindly: These now need to be stdout

cove_iati/management/commands/iati_cli.py
```diff
@@ -17,19 +16,28 @@ def add_arguments(self, parser):
...
             sys.exit(1)
         except CoveInputDataError as e:
             self.stdout.write('Not well formed XML: %s' % str(e.context['error']))
```
kindly: stdout also.

We should probably move both these exceptions up to wrap ```iati_json_output(..)```
As if one file fails we probably want a way for the streaming to continue.
We could have a special json object in the stream like
```{"file": "/path/to/file", "error": "what the error is"}```

File: cove_ocds/management/commands/ocds_cli.py [hunk 17]
```diff
@@ -30,10 +31,18 @@ def handle(self, file, *args, **options):
             ))
 
         try:
-            result = ocds_json_output(self.output_dir, file, schema_version, convert)
+            if stream:
+                for file in files:
+                    if os.path.isdir(file):
+                        self.stdout.write('Skipping %s directory ' % str(file))
```
kindly: stderr again

cove_ocds/management/commands/ocds_cli.py
```diff
@@ -30,10 +31,18 @@ def handle(self, file, *args, **options):
...
+                    else:
+                        result = ocds_json_output(self.output_dir, file, schema_version,
+                                                  convert)
+                        self.stdout.write(str(result))
```
kindly: Same comments that apply to IATI apply here.

cove_iati/tests.py
```diff
@@ -220,6 +220,22 @@ def test_cove_iati_cli_delete_option():
...
+    call_command('iati_cli', file_path, stream=True, output_dir=output_dir)
+    assert not os.path.exists(output_dir)
+
+
+def test_cove_iati_cli_stream_option_dir():
+    file_path = os.path.join('cove_iati', 'fixtures')
+    test_dir = str(uuid.uuid4())
+    output_dir = os.path.join('media', test_dir)
+    call_command('iati_cli', file_path, stream=True, output_dir=output_dir)
+    assert not os.path.exists(output_dir)
```
kindly: We need some tests that actually show multiple files being used and also some test actually looking at the output.

cove_iati/management/commands/iati_cli.py
kindly: We also should add the filename to the ouput JSON also.

ghost:
Closing due to age and as OCDS is now in a different repository. This feature should live in https://github.com/open-contracting/lib-cove-ocds now.

Issue #915, opened 2017-12-12
Add mercurial to the requirements for IATI localhost


rory09:
I've just been following the instructions to install IATI CoVE to run on localhost so that I can use the CLI, and I've hit an issue with a dependency on @Bjwebb's bitbucket account:

```
Collecting defusedxml==0.5.0 (from -r requirements_iati.txt (line 4))
  Downloading defusedxml-0.5.0-py2.py3-none-any.whl
Obtaining defusedexpat from hg+https://bitbucket.org/bjwebb/defusedexpat@11b0ba9f42a547607134f7b36a241382659aff24#egg=defusedexpat (from -r requirements_iati.txt (line 5))
  Cloning hg https://bitbucket.org/bjwebb/defusedexpat (to revision 11b0ba9f42a547607134f7b36a241382659aff24) to ./.ve/src/defusedexpat
  Error [Errno 2] No such file or directory: 'hg' while executing command hg clone --noupdate -q https://bitbucket.org/bjwebb/defusedexpat /home/rory/Code/cove/.ve/src/defusedexpat
Cannot find command 'hg'
```

I fixed this by installing mercurial. I'd recommend adding it to the installation guide assuming I didn't just miss something?

Bjwebb:
You're quite right, it's just missing from the installation guide. Here's a PR to fix that - https://github.com/OpenDataServices/cove/pull/923

andylolz:
Hmmmmm are you aware of this? https://github.com/tiran/defusedexpat/commit/f16cbaeee2595c0f30eeee1292b97f1557ddb43f

Bjwebb:
We're using Python 3.5, so it's not a simple matter of just switching to that repo.  I will take a look at whether we can build on top of the github repo though.

andylolz:
> We're using Python 3.5, so it's not a simple matter of just switching to that repo

Apologies! I looked at your bitbucket pull request (about python3.4) rather than your latest bitbucket commit. My mistake!

That said, the latest commit looks pretty trivial. So while it’s not a simple matter of just switching to that repo… I wonder if it is a marginally-less-simple-but-still-pretty-simple matter of switching to a github fork with one extra commit?

Bjwebb:
@andylolz Thanks for the suggestion, we're building on top of the GitHub repo now.

PR #923, opened 2017-12-18
[#915] Add mercurial to IATI CoVE installation requirements

Bjwebb:
#915

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,7 +113,7 @@ The following steps are for Ubuntu but equivalent packages are available for oth
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


edugomez:
@Bjwebb have you consider https://github.com/OpenDataServices/cove/issues/915#issuecomment-352523907 which would avoid a new dependency?

Bjwebb:
We're using Python 3.5, so it's not an easy swtich. I will look into it, but suggest we merge this in the meantime. (Apt dependencies generally want a review, as we should be installing wheels, which removes the need for libxml2-dev etc.)

Issue #921, opened 2017-12-18
Add information about the behaviour of cove_ocds shared URLs  


edugomez:
We need to clarify the state of shared Cove data URLs  

When users upload data convert data, results and the conversion are modified if they switch to a different version of the schema. When sharing the URL for their data, the receiver will see that state of the validated data in Cove.

edugomez:
We decided to leave the current behaviour as it is, adding a tooltip to inform users of cove behaviour when an URL is shared  https://github.com/OpenDataServices/cove/pull/924

PR #924, opened 2017-12-18
[#921] Add db field data_version_schema and CSS tooltip

edugomez:
#921 

Also, addition of an extra db field (data_version_schema, so it contains [**a migration**](https://github.com/OpenDataServices/cove/pull/924/files#diff-687ed75c63d2f1439a9a5f724814043f) ) to store version in data (1.1 and above) and display in headlines.
926_a


edugomez:
screenshoot_tooltip

Issue #925, opened 2017-12-19
'list' object has no attribute 'values'


edugomez:
https://sentry.io/open-data-services/cove-dev/issues/420763269/

```
AttributeError: 'list' object has no attribute 'values'
(2 additional frame(s) were not displayed)
...
  File "cove/lib/exceptions.py", line 43, in wrapper
    return func(request, *args, **kwargs)
  File "cove_iati/views.py", line 86, in explore_iati
    context = common_checks_context_iati(context, db_data.upload_dir(), data_file, file_type)
  File "cove_iati/lib/iati.py", line 117, in common_checks_context_iati
    for rules in ruleset_errors.values():

AttributeError: 'list' object has no attribute 'values'
```

Bjwebb:
I've approved the pull request, so we can get the fix live ASAP.

It would be good to also see a comment in the code about what's going on, and ideally a test.

edugomez:
I was thinking of looking in more depth into it here https://github.com/OpenDataServices/cove/issues/927 (open)

PR #926, opened 2017-12-19
Check whether ruleset errors is a list (CLI) or dict (web)

edugomez:
#925


edugomez:
@Bjwebb quick note on this bug.  ```get_iati_ruleset_errors``` returns either a list or a dict, which I think it is messy and prone to bugs like the one in this issue (@kindly was working on the JSON output whilst I was working on by rule/by activity grouping, so I had to be careful not to mess with the list output). Not typing in Python can be quite handy but you pay a price for it!

Issue #911, opened 2017-12-11
Cache schema and extensions


jpmckinney:
I described to @timgdavies that I was using CoVE as a library #908 to validate multiple packages, and he mentioned there may be a performance issue, due to CoVe re-downloading schema and extensions for each package. If that's the case, could we cache those at runtime?

robredpath:
Talking to the team today, we'd like to profile performance first - just to check that it is a reasonable place to look. We'd also want to make sure that we fully understand how this might affect the web version - we think we do, but we want to be careful. So, our plan is to do that, then do this caching if it looks like it will help....or do something else if that'll have more of an impact.

robredpath:
@jpmckinney - can you share the data that you're using for this?

jpmckinney:
Here's a script I'm running for the root of the cove directory:

```
import os
import sys
from os.path import basename

import requests

os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cove_ocds.settings")

from cove_ocds.lib.api import APIException, ocds_json_output

page = 1

while True:
    response = requests.get('https://www.contractsfinder.service.gov.uk/Published/Notices/OCDS/Search?page='.format(page)).json()
    for result in response['results']:
        # Consider combining big release package per page of results
        with open('data.json', 'w') as f:
            f.write(requests.get(result['uri']).text)

        try:
            output = ocds_json_output('', 'data.json', '', False)
            if output['validation_errors']:
                print(result['uri'], repr(output['validation_errors']))
                sys.exit(1)
            else:
                print('.', end='', flush=True)
        except APIException as e:
            print(str(e))
            sys.exit(1)

    if page == 4:
        break

    if response['maxPage'] == page:
        break
```

robredpath:
We're implementing this using https://docs.python.org/3.5/library/functools.html#functools.lru_cache 

We've not looked at performance yet, as this is an easy implementation, so might as well do it first anyway!

edugomez:
The script above with no schema/extensions caching:
```bash
........................................................................................https://www.contractsfinder.service.gov.uk/Published/Notice/releases/9383b4d8-b69d-4dbe-a382-f235e4675c26.json [{'value': '', 'path': 'releases/0/awards/0/suppliers/0/name', 'description': "'' is too short", 'type': 'minLength', 'field': 'releases/awards/suppliers/name'}]

real    2m17.313s
user    0m22.668s
sys     0m0.520s
```
With caching:
```bash
[...]

PR #932, opened 2017-12-22
911 cache schema

edugomez:
#911


cove_ocds/management/commands/ocds_cli.py
```diff
@@ -13,15 +13,20 @@ class Command(CoveBaseCommand):
...
 
     def add_arguments(self, parser):
-        parser.add_argument('--schema-version', '-s', default='', help='Version of the schema to validate the data')
-        parser.add_argument('--convert', '-c', action='store_true', help='Convert data from nested (json) to flat format (spreadsheet) or vice versa')
+        parser.add_argument('--schema-version', '-s', default='',
+                            help='Version of the schema to validate the data')
+        parser.add_argument('--convert', '-c', action='store_true',
+                            help='Convert data from nested (json) to flat format (spreadsheet) or vice versa')
+        parser.add_argument('--cache', '-z', action='store_true',
+                            help='Cache the remote schema and extensions in memory')
```
kindly: I would say by default the cli should cache. So I am not sure there is a need for that argument.
We could have an option --nocache but not sure its necessary.

edugomez: Will change to default caching.

> We could have an option --nocache but not sure its necessary.

Prob not necessary.

PR #953, opened 2018-01-24
Fix IATI translations by removing Plural-Forms line

Bjwebb:
Also remove an empty translation from a template.


robredpath:
@Bjwebb how can I test this? Do I need to before approving?

Bjwebb:
If you run `DJANGO_SETTINGS_MODULE=cove_iati.settings python manage.py runserver` on this branch, and check to see if it's not broken.

PR #958, opened 2018-01-31
Add OCDS-specific Ts&Cs

robredpath:
For avoidance of confusion (mostly my own!) over template overriding, and to avoid accidental changes via overriding, I've copied the Ts&Cs into their own file and modified just for OCDS. We may want to apply the same change to 360, but I'm keen to get this rolled out for OCDS so that I can look at uploads for a week after the change is applied.


Bjwebb:
Could you split this into two commits, one copying the Ts&Cs, and another making the change?

robredpath:
@Bjwebb done ^^

Issue #675, opened 2017-05-10
OCDS: Check ocid prefixes


timgdavies:
We should have introduce an ```ocid``` prefix check during validation of OCDS data.

duncandewhurst:
+1 to this - feedback from OCP is that it doesn't make sense for a file to validate when no ocid prefix is used

robredpath:
Is the ask here for a check against the register of ocids or a check that the ocid is of a valid form? @timgdavies @duncandewhurst ?

duncandewhurst:
Both please, so a check first that it is of valid form and a check second against the ocid prefix google sheet (checking against the CSV version so that in future a different CSV could be substituted if we move to a different registration method)

robredpath:
Thanks @duncandewhurst . Is it still useful with just a format check? Talking to an external (hosted) list is a new step that needs quite a bit of thinking about to do well (eg what do we do if it's down? slow? misformatted? do we cache? ) . We can do it, but it'll push something else back, so trying to gauge priority

timgdavies:
This could definitely be split into two parts, with a format check, and then later on a list check.

robredpath:
Prep required is to get agreement that a format check is a sufficient first step

robredpath:
We're thinking of starting a warnings box (probably not called that), in which this would be a useful first entry.

robredpath:
We still need to decide on the name of the box. In #947 I suggested 'Quality Factors', but I'm not sure.

edugomez:
What about 'Quality Warnings', which will have 4 headers:

_Field - Value - Warning - Path to field_

For ocds prefixes it will be e.g:

_ocid, ocds-bad-123456, The Open Contracting ID contains a malformed OCDS prefix, releases/0/ocid_

robredpath:
@edugomez let's give it a try and see what people think! If you could post some screenshots we can get a couple of friendly users to take a look and see what they think

edugomez:
```id```s in objects in the standard can use the ```ocid``` as its value, so if the prefix in ```ocid``` is wrong  it will wrong in the ```id``` too. Do we need to check for ```id```s prefixes too?

Bjwebb:
I would think just checking the `ocid` for now should be fine.

timgdavies:
I'm working on a language review across the validator, but I *think* we would want to put these under a conformance heading, rather than quality.

[...]

PR #963, opened 2018-02-02
Check OCDS Prefixes

edugomez:
#675


edugomez:
Some comments on this PR:
- No translations yet
- It will probably conflict with the code in this other PR https://github.com/OpenDataServices/cove/pull/967
- You can see some screenshots in the corresponding issue #675
- It could do with a more detailed functional test. At the moment it checks for the section 'Conformance' and the message with the specific number of fields identified as having bad ```ocid``` prefixes, but it doesn't click on _View all errors_ to inspect the actual table with ```ocid```s and paths.

PR #964, opened 2018-02-02
Fix terms_ocds URL

robredpath:


cove_ocds/urls.py
```diff
@@ -13,9 +13,9 @@
 urlpatterns_core += [url(r'^data/(.+)$', cove_ocds.views.explore_ocds, name='explore')]
 
 urlpatterns = [
+    url(r'^validator/terms_ocds', TemplateView.as_view(template_name='terms_ocds.html'), name='terms'),
     url(r'^$', RedirectView.as_view(url='validator/', permanent=False)),
     url('^validator/', include(urlpatterns_core)),
-    url(r'^validator/terms_ocds', TemplateView.as_view(template_name='terms_ocds.html'), name='terms'),
```
edugomez: @robredpath What's the difference here?

robredpath: @edugomez I thought that Django used the last matching pattern to resolve URLs, turns out it stops after the first matching pattern. Therefore, I've moved it to the top of the list so that the matching pattern in urlpatterns_core doesn't get matched.

edugomez: @robredpath yes, but there isn't a ```r'^validator/terms_ocds'``` in ```urlpatterns_core``` so I wonder what was the problem in the first place. If I go to _http://standard.open-contracting.org/validator/terms_ocds_ I can see the OCDS T&C with no problem. 

I may be missing something, but the only potential problem I can see here is having two ```r'^$'```. However, the ocds one comes first, that's why you get redirected to ```/validator``` and index of core is not shown again.

robredpath: @edugomez if you go to http://standard.open-contracting.org/validator/terms_ocds then you see the old terms - look for the word 'review' on that page, and you won't find it. 

In urlpatterns_core there's a pattern that matches any URL that starts with /terms, and it's not terminated - so it matches /terms_ocds as well as just /terms. So, this provides a more specific URL that'll match first.

edugomez: > In urlpatterns_core there's a pattern that matches any URL that starts with /terms,

Ah, OK, sorry, I missed the lack of ```$```. What is the intention behind a URL that matches any  _/terms*_. I think that having to rely on keeping track of what URL Django will find first is not a good strategy. Much better to be deterministic, e.g. create a T&C for every cove, if they are waiting forT&C to be created of they will generic / shared, redirect to the core T&C.

edugomez: @robredpath to be clear, I will add ```'$'``` to the core _/terms_ URL

robredpath: @edugomez this is the first time that anyone's created a custom Ts&Cs, and it's likely to be a temporary thing anyway. But, do go ahead and add that $, it's a good thing to fix anyway.

Bjwebb:
@robredpath Could you double check my commit looks good. Then, once tests pass I can get this merged and deployed.

robredpath:
@Bjwebb the changes look good, happy for this to be merged once tests pass

Bjwebb:
Now at http://standard.open-contracting.org/validator/terms_ocds/

PR #970, opened 2018-02-09
Get rid of intermittently failing assertion in OCDS schema cache test

edugomez:
#970

### DIFF
--- a/cove_ocds/tests.py
+++ b/cove_ocds/tests.py
@@ -1012,7 +1012,7 @@ def test_cove_ocds_cli_schema_cache():
         cache_time = cache_end - cache_start
         cache_time_total += cache_time
 
-    cache_time_avg = cache_time_total / 5.0
+    cache_time_avg = cache_time_total / 6.0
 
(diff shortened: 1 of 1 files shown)


cove_ocds/tests.py
```diff
@@ -1003,16 +999,7 @@ def test_cove_ocds_cli_schema_cache():
     assert sorted(results['extensions']['invalid_extensions'], key=lambda k: k[0])[1][1] == '404: not found'
 
     # Use cached schema from previous call_command
-    cache_time_total = 0
-
-    for i in range(6):
```
Bjwebb: I'd quite like to keep this test if we can.

I note that `range(6)` has 6 items `[0, 1, 2, 3, 4, 5]`, but the average below is calculated by dividing by 5. Perhaps aligning these will be sufficient to stop the test failing?

edugomez: @Bjwebb well spotted! That may be it. I'll fix it.

PR #973, opened 2018-02-13
Pass lxml a raw file handle so that it can detect the encoding

Bjwebb:


Bjwebb:
This fix now on http://iati.cove.opendataservices.coop/

PR #975, opened 2018-02-14
Remove broken google doc functionality

Bjwebb:
This ensures that we can use XLSX links from google sheets, without the special handling doing the wrong thing.

Dev deploy link http://remove-google-doc.cove-360-dev.default.threesixtygiving.uk0.bigv.io/


Bjwebb:
Now on http://master.cove-360-dev.default.threesixtygiving.uk0.bigv.io/

Bjwebb:
This is also now deployed to http://iati.cove.opendataservices.coop/

PR #983, opened 2018-02-21
952 favicon font

kindly:


Bjwebb:
The text looks a bit too small with this font. Is it possible to deal with that by increasing the font size slightly?

kindly:
Yes I thought it was small too.

It is hard to just change the font size of just say body/paragraph text in a way that bootstrap themes like, to make everything (like line spacing) consistent.  They are all relative to a base font size.

I have increased the base font size in the latest commit but it does increase the sizes of lots of other elements too.  However, I think I like that.

Bjwebb:
I'm pretty happy with this fontsize, but it was looking different on my high res screen laptop (now in for repair). @robredpath Do you have a "retina" screen? Could you test this for us?

robredpath:
I've had a look at this on my Retina display, both with default scaling and at native resolution, and it looks 'right' to me in both cases.

Bjwebb:
Deployed to http://standard.open-contracting.org/validator/

PR #984, opened 2018-02-26
Update start date of new Ts&Cs

robredpath:


Bjwebb:
As you've noticed this lacked translations, which I've now added - I think I trust google translate for a simple date!

Issue #959, opened 2018-01-31
IATI organisation file support


siemvaessen:
We are looking into CoVe for converting IATI organisation file types, next to the activity file. 

We don't see support for this in CoVe (yet?) so we are wondering what your thinking is on implementing such a feature and if you do not have short term plans to do so, we'd be happy to talk about how we could contribute to this.

Bjwebb:
Yes, support for this doesn't exist yet, and it makes sense to implement this in CoVE.
Our conversion library flatten-tool nearly supports this, it's just missing a way to name the root XML element (`iati-organisations`). There's an issue for this https://github.com/OpenDataServices/flatten-tool/issues/150

There's also some discussion in this issue - https://github.com/OpenDataServices/cove/issues/732 - about how CoVE might auto detect organisation spreadsheets from the Meta tab.

We have no short terms plans to do anything on this, so would by happy to see you contribute.

siemvaessen:
Thanks @Bjwebb. On the library flatten-tool, any ETA on delivery? 

And, where should we look for relevant code in CoVe on the org file detection yet?

Bjwebb:
Sorry for the delay on this, I had a long weekend away at FOSDEM.

We don't have an ETA on that flatten-tool issue atm, but it should be fairly straightforward, so something we can slot into our backlog. How quickly are you hoping to get this functionality?

For the CoVE the changes need to happen in a few places. I'm thinking a call might be useful to talk over these?

siemvaessen:
Hi @Bjwebb - delay in answering. Will CoVE support this in Q2 this year? We'd be happy to contibute, but I'd first like to find out where CoVE now stands with org.file support. We could on the base what is out there continue building.

Bjwebb:
We don't currently have this on our roadmap, but we'd definitely be happy to support you to do this.

siemvaessen:
check. Closing this ticket then.

andylolz:
Any chance of reopening this issue? I’d also find org file support useful.

andylolz:
Thanks for reopening, @Bjwebb!

> we'd definitely be happy to support you to do this.

I don’t know how far this invitation extends?! But OpenDataServices/flatten-tool#201 is my first blocker. Can I get a sense of whether / when it will be reviewed / merged? Thanks!

Bjwebb:
Thanks, I'll try and take a look at that this week.

siemvaessen:
[...]

PR #993, opened 2018-03-20
Validate against IATI org schema

### LINKED ISSUE #922: schema_aiti typo
Should be schema_iati

andylolz:
Make a start at #959.

Also happens to fix #922.


Bjwebb:
Thanks for taking a look at this.

Am I right in thinking this validates against the org schema first, and if that fails tries the activity schema instead? I suspect this would be somewhat confusing behaviour.

andylolz:
> Am I right in thinking this validates against the org schema first, and if that fails tries the activity schema instead? I suspect this would be somewhat confusing behaviour.

That’s correct, yes. This PR isn’t a solution – as described, it just makes a start at tackling the problem.

So you’ll see a failure against the activity schema if you try to test an invalid org file. I agree that this would be confusing (though note that this is what happens currently anyway.)

I guess the correct approach would be to peek at the data and attempt to identify which schema to use?

andylolz:
> peek at the data and attempt to identify which schema to use

I did it this way, and added a test. @Bjwebb does this look okay?

andylolz:
This issue is also dependent on OpenDataServices/flatten-tool#150.

andylolz:
Cool – a load of stuff that was blocking this has now been cleared!

I have a nagging feeling that I wanted to get the meta tab roundtripping before finishing this one off, but I don’t exactly remember why…

Bjwebb:
Before this Pull Request can be merged, I think the remaining tasks are:
- [ ] Create a new tagged version of flatten-tool
- [ ] Update the requirements files in this pull request
- [ ] Someone who's not me to review the code

andylolz:
:clap::clap::clap:

Issue #851, opened 2017-09-28
"Sorry we can't process that data" is hidden


Bjwebb:
e.g. 

![image](https://user-images.githubusercontent.com/634/30976553-012a1c7c-a46d-11e7-85b4-a01266b344b0.png)

![image](https://user-images.githubusercontent.com/634/30976577-0fbb2998-a46d-11e7-900c-3461594256f0.png)

robredpath:
@Bjwebb is this the same for all CoVEs?

Bjwebb:
No, looks like it's only a problem for the IATI CoVE.

andylolz:
See: https://github.com/thomaspark/bootswatch/issues/341

Bjwebb:
@andylolz There's a commit on that upstream issue that claims to fix it, and should be included in the sandstone version we're using. But, it's obviously still broken for us. Do you have an idea as to why that is?

andylolz:
Yeah – it’s because it doesn’t fix it at all.

andylolz:
Yeah – that’s what I was referring to when I said it seems doubtful it will be fixed upstream (because the owner closed the issue with a bad fix, but also because IIRC the repo is mostly about bootstrap4 now).

I did look into this at the time, and concluded the fix was bad. I should have noted why, because I can’t remember now. Glancing at it again, I guess the fix sets the text colour on `.panel .panel-heading`, but that gets overridden by the more specific selectors.

PR #994, opened 2018-03-20
Hack a fix for theme bug (because it seems doubtful this will be fixed upstream).

andylolz:
Refs #851.


andylolz:
NB this requires a hard refresh because there’s no asset fingerprinting.

PR #995, opened 2018-03-20
Make footer full width

andylolz:
I guess container should be inside footer, not the other way around.

### DIFF
--- a/cove/templates/base.html
+++ b/cove/templates/base.html
@@ -108,8 +108,8 @@
 
   </div><!-- bottomcontent  -->
 
-  <div class="container">
-    <div id="footer">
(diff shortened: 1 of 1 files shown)


Bjwebb:
What's the purpose of this? AFAICT it just makes the horizontal line full page width, which is inconsistent with the line below the heading:
Current master:
image
This PR:
image

andylolz:
IAMAD (I Am Not A Designer) but this is consistent with the way bootstrap does it:
https://getbootstrap.com/docs/3.3/examples/sticky-footer/

screen shot 2018-03-27 at 10 23 42

…although not with the way bootswatch does it.

So I’ll close this.

Issue #846, opened 2017-09-21
OCDS: Fix mismatching greens


duncandewhurst:
I just noticed there are three different greens used for the boxes on the validator results page.

I think they should all be the same as the green used for the headlines box.

![image](https://user-images.githubusercontent.com/19265814/30695778-afc4ee4a-9ed9-11e7-8b15-f26b3c0098f1.png)

andylolz:
So:
 * Headlines, Documents and Key Field Information all use `panel-primary`
 * Schema uses `panel-warning`
 * Convert uses `panel-success`

IATI and 360 use `panel-default` everywhere, which I suppose is preferable (for default stuff, that isn’t an error or a warning).

PR #1000, opened 2018-03-21
Use `panel-primary` as the default panel class (OCDS)

andylolz:
Refs #846. It would probably be preferable to use `panel-default`, but I’ve done it this way to make the PR a bit simpler.


Bjwebb:
@kindly Could you review this one please? You're more familiar with the theme than I am.

PR #1001, opened 2018-03-26
Hack a fix for theme bug (because it seems doubtful this will be fixed upstream).

Bjwebb:
Copy of https://github.com/OpenDataServices/cove/pull/994


andylolz:
You have The Power™ to commit to my PRs (if that helps).

PR #1003, opened 2018-03-27
Update requirements

Bjwebb:


cove_iati/lib/exceptions.py
```diff
@@ -8,7 +8,7 @@ def __init__(self, context, errors=''):
         self.feature_name = context.feature.name
         try:
             self.id = context.xml.xpath('iati-identifier/text()')[0]
-        except:
+        except Exception:
```
andylolz: Presumably this is an IndexError?

andylolz: (i.e. presumably you can catch a more specific error, which I think is what flake8 is encouraging you to do here).

andylolz: (I.e. I wonder if you could be more specific when catching this and other exceptions – particularly given this commit is called “Fix flake8”, and I think PEP8 says you should aim to do this.)

Issue #998, opened 2018-03-20
What should “IATI Schema” link to?


andylolz:
Just here:
https://github.com/OpenDataServices/cove/blob/9072e97265142c6167f56d917b3cff2bbc6887b6/cove_iati/templates/cove_iati/explore.html#L112

I guess it should at least state the version (currently v2.02) and perhaps link to http://iatistandard.org/202/schema/ or something.

Also, why isn’t “IATI Schema” translated?

Bjwebb:
Thanks. Yes, I think this is right, and we're now 2.03.

PR #1004, opened 2018-03-28
Add link to IATI Schema page

andylolz:
Refs #998.

“IATI Schema” should be translated, too, but that’s a separate issue (and I’m not totally sure I’d get it right!)


cove_iati/templates/cove_iati/explore.html
```diff
@@ -109,7 +109,7 @@ <h3 class="panel-title panel-title-explore">
         {% else %}
           {% trans "Congratulations! Your data is valid against" %}
         {% endif %}
-          <a href="http://iatistandard.org/203/schema/">IATI Schema</a>.
+          <a href="http://iatistandard.org/203/schema/">IATI Schema version 2.03</a>.
```
andylolz: Oh yes – good point!

Issue #940, opened 2018-01-08
'None' in headings warning should perhaps be suppressed


rory09:
When I model data a google sheet, download the file and uploaded it to IATI CoVE, I occasionally get this conversion error (see [here](http://iati.cove.opendataservices.coop/data/2d8d2b0b-c80a-408c-bb0e-935d2acfe29f)):

"Duplicate heading "None" found, ignoring the data in column"...

![image](https://user-images.githubusercontent.com/7805169/34671606-2976d37c-f473-11e7-9c21-46be420bd5cf.png)

This tends to be when I have populated columns, but then rearranged things (i.e. I had many columns on one sheet but then moved transactions to their own sheet) leaving empty values like this:

![image](https://user-images.githubusercontent.com/7805169/34671641-414021ca-f473-11e7-8550-86b26ef74a1a.png)

This doesn't make the conversion fail, and the XML output is still valid, but it still shows up as an *error* on CoVE's UI.

For the following reasons, I think we should suppress this:

* It's misleading, and could make a user try and 'fix' the error, which AFAIK they can't and shouldn't do
* Again, AFAIK, I don't think this can easily be fixed, and the way I can think of isn't reasonable (actually deleting the columns in Google Sheets or something similar
* It's not like another duplicate header - it's a 'None' value in a header, which, unless there are values below it, should be assumed to be empty and not of interest in conversion.

Thanks, R

Bjwebb:
Agreed, those messages aren't useful.

rory09:
Heya - just bumping this. I'm not sure if this has made it into another sprint plan, or if we could maybe put it into the non-sprint board? cc @robredpath @Bjwebb

robredpath:
@Bjwebb any sense of scale on this?

Bjwebb:
~1 hour

Bjwebb:
Sorry, looks like I failed to update the flatten-tool commit CoVE uses correctly (I missed `requirements.in` in https://github.com/OpenDataServices/cove/commit/d9e784e6484792a9363bdbfa1ae45907b71f73f7), so this change has been reverted by accident. I will fix by setting the commit to the correct value again.

andylolz:
> so this change has been reverted by accident

It looks like it was reverted as part of #972, and it looks like it happened **after** the review (during the bit that says “I'm working through the merge conflicts”).

Given that merge conflicts are sometimes arduous, I wonder if it would be prudent to (re-)review **after** these have been resolved?

[...]

PR #1006, opened 2018-03-28
[#940] Update flatten-tool

Bjwebb:
#940 

Only change to behaviour is to remove duplicate None warnings during Spreadsheet conversion
https://github.com/OpenDataServices/flatten-tool/pull/199

This will affect all instances of CoVE.


Bjwebb:
Impact of this pull request on 360Giving is that Wellcome's data no longer has a conversion error:

Before:
image

After:
image

@KDuerden Does that look okay?

KDuerden:
Thanks for this. Good to go from my perspective!

PR #1009, opened 2018-04-16
 Update OCDS CoVE to 1.1.3

Bjwebb:


Bjwebb:
Yes, the test updates relate to changes in the schema, that get pulled in automatically, so are required for the tests to pass.

PR #1010, opened 2018-04-18
Changes to handle 360Giving standard date changes better

Bjwebb:
* Remove date format workaround
* Reinstate old date validation error message (and group correctly, and keep the link to the common errors page)
* Update the tests


Bjwebb:
Here's a deploy to test this: http://date-validation.cove-360-dev.default.threesixtygiving.uk0.bigv.io/
@KDuerden: I think this covers everything we talked about, apart from changing the text of the common errors page.

KDuerden:
I've tested with various bits of not quite right dates and this is working well. Happy with this, thank you very much! I'll look at the common errors text in the light of this update.

Issue #704, opened 2017-06-12
Extensions list in CoVE should link to documentation as well as extension JSON


timgdavies:
At the moment the list of Schema Extensions provides links to the release-schema.json for each extension. 

![image](https://user-images.githubusercontent.com/342624/27021807-b20741da-4f42-11e7-9f67-9e4a7c6b1872.png)

I though we were going to link this to the **documentation URL** from extension.json to give users a link to a clear explanation of the extension they are using. Perhaps this could be paired with an 'extension source' link to the release-schema.json.

![image](https://user-images.githubusercontent.com/342624/27021902-339b3b48-4f43-11e7-96b1-aa4835bfc1db.png)

However, we should also be careful to note that extensions may patch release-package-schema.json or record-package-schema.json. Have we got support for this right now? If not - we should open a separate issue on that.

timgdavies:
This is a blocker for us blogging about extensions, which we want to do during June.

robredpath:
@timgdavies do you have a separate piece of work planned to add the documentation URL to extension.json for all of the extensions? The lots extension, for example, doesn't have that - unless I'm looking in the wrong place? https://github.com/open-contracting/ocds_lots_extension/blob/master/extension.json is what I'm looking at. 

We will add the function to CoVE to pull out the documentation URL from extension.json where it is supplied, and link to it. What will the element be called?

timgdavies:
Ah: I didn't realise that was dependency here. I had forgotten we were not using the extension registry to capture this information. 

Ok: I think we might need quite a strong extension focus in next OCDS sprint to get this done - as we'll need to update extension templates and extension creator accordingly. 

Where best to start putting together a cross-project backlog for all of this?

robredpath:
@timgdavies the developers are meeting this afternoon to see what the work for making the tools we maintain work for 1.1 looks like. Once we've got that, we can have a chat about bringing together some of that work with some extensions work. 

I've put a call in for the two of us on Thursday afternoon - I think if you come to that with a list we'll be able to work it out from there!

duncandewhurst:
I just noticed that the validator doesn't seem to be doing this, [for example](http://standard.open-contracting.org/validator/data/d3817878-b147-4a46-a1f2-70840c64bc19) these two links point to the release-schema.json files for the two extensions:

![image](https://user-images.githubusercontent.com/19265814/41382429-3c429dce-6fc0-11e8-872f-4185d04091cd.png)

[...]

PR #1026, opened 2018-06-15
release-schema.json is optional for extensions

### LINKED ISSUE #1024: OCDS: release-schema.json is optional
An extension may only extend one of the package schemas, or only codelists. Presently, CoVE will fail with 'fetching failed'.

rhiaro:
Fixes #1024 and #704, adds a test for an extension with no release-schema.json, and updates error handling so existing tests still pass.


Bjwebb:
This is now live on http://standard.open-contracting.org/validator/

Issue #1022, opened 2018-05-24
Show dates in the interface in a human friendly format


duncandewhurst:
Dates in the **Key Field Information** and **Release Table** sections of the validator results are currently shown in ISO8601 format:

![image](https://user-images.githubusercontent.com/19265814/40462985-a3e0553a-5f66-11e8-9a25-704015cf95ed.png)

Since I assume we're only displaying valid dates there anyway and in the spirit of using CoVE to explore data, it would be nice to show those in a human friendly format.

robredpath:
+1 - I'd be happy for us to do this. It'll require a bit of thought on the design (how to show timezones, for example) but I think is pretty straightforward to do.

andylolz:
> how to show timezones, for example

Maybe hide time and timezone info? Looks like a bunch of stuff in the standard is documented as a date, but is represented by a date-time internally. Perhaps the interface should reflect the way it’s documented, rather than the internal representation. Idk

PR #1031, opened 2018-06-21
Human-friendly dates in frontend

rhiaro:
Something went horribly wrong with updating the requirements.txt, but otherwise this adds human friendly (`1 Apr 2014, 12:00 (UTC)`, including month name localisation courtesy of django) dates per #1022


Bjwebb:
> Something went horribly wrong with updating the requirements.txt

Bjwebb:
> Something went horribly wrong with updating the requirements.txt

By default `./update-requirements.sh` updates all requirements to the latest version. `./update-requirements-sh --new-only` will just add new requirements for new entries in requirements.in

cove_ocds/views.py
```diff
@@ -162,6 +163,16 @@ def explore_ocds(request, pk):
         template = 'cove_ocds/explore_release.html'
         if hasattr(json_data, 'get') and hasattr(json_data.get('releases'), '__iter__'):
             context['releases'] = json_data['releases']
+
+            # Parse release dates into objects so the template can format them.
+            for release in context['releases']:
+                if hasattr(release, 'get') and release.get('date'):
+                    release['date'] = parser.parse(release['date'])
```
Bjwebb: `dateutil`'s parser is great because it is accepts dates in a lot of different formats. However, where we're promoting a data standard, I think there's an argument that we should be stricter with how we parse dates, such that we don't accept bad data. (OTOH, the validation info will still be there).

Tagging others for their thoughts on this: @kindly @robredpath @timgdavies

Bjwebb: Also looks like we can get a ValueError here, which is why the tests are failing. Here's an example file: https://raw.githubusercontent.com/OpenDataServices/cove/master/cove_ocds/fixtures/badfile_all_validation_errors.json

robredpath: Oh, well this is a fun question. I could make the arguments either way, but I think I'd be inclined to say that:
- We shouldn't pretend that data's ok if it's not valid
- But, CoVE is jam-packed with things that make people think about their data, and one of those things is that we show their data back to them to make them see if it's wrong
- So, could we put some kind of warning if dates aren't valid, but make our best guess as to what they are? Maybe show it in red, and say "unknown - best guess XXXX-XX-XX - click here to find out more"?

timgdavies: ## Checking understanding

As I understand the issue:

We want to display dates to be people in localised formats, so that instead of 

```
Earliest release: 2018-06-05T21:35:12.323Z
Latest release: 2018-06-08T11:27:30.543Z
```

we having something _like_:

```
Earliest release: 05 Jun 2018 at 21:35
Latest release: 08 Jun 2018 at 11:27
```

Someone may provide an invalid date format (e.g. `"date":"2018/06/02"`)

This will still fail validation and generate a validation error

But, in parsing it with a forgiving library, we can still display:

```
Earliest release: 05 Jun 2018 at 21:35
Latest release: 08 Jun 2018 at 11:27
```

with the risk that (a) we're hinting to them that their data was ok really ("If CoVE could understand it, do I really need to fit it?", and (b) we could parse it wrong and give the wrong answer.

## Thoughts

If we do use a forgiving parser on the date, we should indicate in the UI that we've done this. E.g.

```
Earliest release: 05 Jun 2018 at 21:35 (i)
Latest release: 08 Jun 2018 at 11:27

(i) Dates were provided in an invalid format. We have extracted a best guess at the dates intended.
```

However, as (a) this feels like extra work to maintain; and (b) the 'sense checking' on the contents of the data that the earliest release / latest release type content provides is essentially a second stage of data Quality Assurance **after** first getting it structurally right, which includes date formats, I would lean towards only parsing dates that are valid against the spec (i.e. ISO dates) at this point.

rhiaro: Ah, I had the impression from the original issue that only valid dates would make it through to this point. I used dateutil because parsing the various valid timezones is kind of a pain with datetime.

rhiaro:
I added the extra validation step on dates that are being passed to the frontend, and just don't display them in the summaries if they're not valid. They're flagged and dumped out as part of the validation errors, so it didn't seem worth repeating this in the informational bits.

PR #1037, opened 2018-07-11
Linking to extended codelists

### LINKED ISSUE #1029: OCDS: Link to codelists does not display extended codelists
For codelist validation errors, the codelist is presented as a link, which goes to http://datapipes.okfnlabs.org/csv/html/?url=https://raw.githubusercontent.com/open-contracting/standard/1.1/standard/schema/codelists/{codelist}.csv

image

This doesn't take account of when codelists have been extended, potentially causing user confusion.

We should allow users to view the extended codelist through this link.

rhiaro:
User interface updates to make it more obvious when extended/modified codelists have been used for validation. Fixes #1029.

### DIFF
--- a/cove/lib/common.py
+++ b/cove/lib/common.py
@@ -800,16 +800,29 @@ def get_additional_codelist_values(schema_obj, json_data):
 
             if path_string not in additional_codelist_values:
                 
-                if schema_obj.extended_codelist_urls and schema_obj.extended_codelist_urls.get(codelist):
-                    codelist_url = schema_obj.extended_codelist_urls[codelist]
-                else:
-                    codelist_url = schema_obj.codelists + codelist
+                codelist_url = schema_obj.codelists + codelist
+                codelist_add_url = None
+                codelist_sub_url = None
+                if hasattr(schema_obj, 'extended_codelist_urls'):
+                    
+                    if schema_obj.extended_codelist_urls.get(codelist):
+                        codelist_url = schema_obj.extended_codelist_urls[codelist]
+
+                    codelistadd = "+" + codelist
+                    if schema_obj.extended_codelist_urls.get(codelistadd):
+                        codelist_add_url = schema_obj.extended_codelist_urls[codelistadd]
+
+                    codelistsub = "-" + codelist
+                    if schema_obj.extended_codelist_urls.get(codelistsub):
+                        codelist_sub_url = schema_obj.extended_codelist_urls[codelistsub]
 
                 additional_codelist_values[path_string] = {
                     "path": "/".join(path_no_num[:-1]),
                     "field": path_no_num[-1],
                     "codelist": codelist,
                     "codelist_url": codelist_url,
+                    "codelist_add_url": codelist_add_url,
+                    "codelist_sub_url": codelist_sub_url,
                     "isopen": isopen,
                     "values": set(),
                     "extension_codelist": codelist not in schema_obj.core_codelists,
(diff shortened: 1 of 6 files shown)


cove/lib/common.py
```diff
@@ -799,11 +799,30 @@ def get_additional_codelist_values(schema_obj, json_data):
             path_string = '/'.join(path_no_num)
 
             if path_string not in additional_codelist_values:
+                
+                codelist_url = schema_obj.codelists + codelist
+                codelist_add_url = None
```
kindly: There is nothing to stop there being multiple add, sub urls. So this should be a list.  I think add/sub should be a single list as order matters.

cove/templates/additional_codelist_values.html
```diff
@@ -15,12 +15,19 @@
       <tr>
         <td>{{detail.field}}</td> <td>{{detail.path}}</td> 
         <td>
-          {% if not detail.extension_codelist %}
-            <a href="http://datapipes.okfnlabs.org/csv/html/?url={{detail.codelist_url}}" >{{detail.codelist}}</td> 
-          {% else %}
-            {{detail.codelist}}
+          <a href="http://datapipes.okfnlabs.org/csv/html/?url={{detail.codelist_url}}" >{{detail.codelist}}</a>
+          {% if detail.codelist_add_url or detail.codelist_sub_url %}
```
kindly: This should be a loop too from the list above.

rhiaro:
I've accounted for multiple codelist modifications in the latest commits. The order the codelist modifications are listed should be consistent with the order extensions are applied. 

On a related note, if multiple extensions *replace* the codelist, the last one to do so is the one that gets linked to, as I think this is the one that is applied. Is that right?

timgdavies:
> On a related note, if multiple extensions replace the codelist, the last one to do so is the one that gets linked to, as I think this is the one that is applied. Is that right?

That seems right to me.

kindly:
@rhiaro yes seems right to me.

jpmckinney:
@rhiaro I can't tell from a quick read, but if an extension adds codes, and then a later extension replaces all codes, will the page still display the `+`, even though that addition became irrelevant?

The following might be used for testing:

```
{
  "version": "1.1",
  "uri": "https://github.com/open-contracting/sample-data/raw/1.1/fictional-example/ocds-213czf-000-00001-02-tender.json",
  "extensions": [
    "https://raw.githubusercontent.com/open-contracting/ocds_location_extension/master/extension.json",
    "https://raw.githubusercontent.com/open-contracting/ocds_tariffs_extension/master/extension.json",
    "https://raw.githubusercontent.com/open-contracting-extensions/ocds_ppp_extension/master/extension.json"
  ],
  "publishedDate": "2010-03-15T09:30:00Z",
  "publisher": {
    "name": "Open Data Services Co-operative Limited",
    "scheme": "GB-COH",
    "uid": "9506232.0",
    "uri": "http://data.companieshouse.gov.uk/doc/company/09506232"
  },
  "license": "http://opendatacommons.org/licenses/pddl/1.0/",
  "publicationPolicy": "https://github.com/open-contracting/sample-data/",
  "releases": [
    {
      "ocid": "ocds-213czf-000-00001",
      "id": "ocds-213czf-000-00001-02-tender",
      "date": "2010-03-15T09:30:00Z",
      "initiationType": "ppp",
      "tag": [
        "tender"
      ],
      "tender": {
        "id": "ocds-213czf-000-00001-01-tender",
        "documents": [{
          "id": "foo",
          "documentType": "foo"
        }]
      }
    }
  ]
}
```

PR #1060, opened 2018-10-17
Fall back on bootstrap default yellow for warnings

### LINKED ISSUE #968: panel-warning in OCDS is green, not yellow
See #947 . We'd like to be able to display warnings, but they come out green, so we resort to errors.

rhiaro:
Fixes #968 subject to shade-of-yellow approval


odscjames:
Looks good, tho I'm a bit confused - is panel-warning used anywhere at the moment? If so, are we happy with the visual changes on the site that will happen? Or is it not used anywhere right now, and this is just setting us up so we can use it in the future?

rhiaro:
Yes, `panel-warning` is used in some of the templates and I definitely managed to generate an error that was affected by this colour change when I was making it.

andylolz:
https://github.com/OpenDataServices/cove/issues/846#issuecomment-375070084 suggests `panel-warning` is used for “Schema” and “Schema extensions” boxes (or at least it was in March).

odscjames:
So I assume we are happy with the visual changes on the site that will happen then?

PR #1085, opened 2018-11-22
Change terminology for validation in OCDS

jpmckinney:
Pursuant to 2944 in CRM.


jpmckinney:
PO files need to be updated (I haven't found instructions, and running `python manage.py makemessages -l es` produces a lot of whitespace differences).

I don't have access to `secret_data_test`, which also causes a failure.

jpmckinney:
This should be part of the Apr-May 2019 sprint.

jpmckinney:
@robredpath Please have this merged early in the sprint – to avoid a conflict.

robredpath:
@kindly will get these changes in at the start of the sprint. This is likely to be through pulling the changes in to a branch owned by an organisational user so that we can sort out the pulling forward of the branch cleanly.

robredpath:
@odscjames will take this on, as this doesn't seem to have happened yet.

odscjames:
This will hopefully be dealt with in https://github.com/OpenDataServices/cove/pull/1164

odscjames:
#1164 has been merged

PR #1090, opened 2018-11-28
Make Cove use lib-cove and lib-cove-web

ghost:


odscjames:
One thing this releases for 360 that wasn't there before is better errors for duplicate IDs: there is an issue for that in this repo somewhere.

cove-360-fix

kindly (approved):

PR #1103, opened 2018-12-07
These packages are no longer in cove (fixes #1102)

### LINKED ISSUE #1102: ReadThe Docs build broken
https://readthedocs.org/projects/cove/builds/8222623/

https://github.com/OpenDataServices/cove/blob/master/setup.py#L20 should not list 'cove' as a package. 

Make sure that doesn't affect anything else then take out!

rhiaro:
Though tbh I'm not sure why `cove_360`, `cove_ocds` and `cove_iati` aren't listed here, but they never were, so..

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ setup(
     version='0.0.0',
     author='Open Data Services',
     author_email='code@opendataservices.coop',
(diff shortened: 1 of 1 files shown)


odscjames:
Before hitting merge, I'm going to get RTD the try and build this branch and make sure that has no problems.

odscjames:
RTD builds fine, thanks!

PR #1125, opened 2019-02-19
Only embedded codelists are checked

### LINKED ISSUE #1124: “Your codelist values are all valid” suggests non-embedded are included
E.g. [this CoVE link](http://iati.cove.opendataservices.coop/?source_url=https%3A%2F%2Fiati-transformer.herokuapp.com%2Ftransform.xml%3Furl%3Dhttps%253A%252F%252Fwww.adb.org%252Fiati%252Fiati-activities-kz.xml). `result/@type` includes codelist values that are invalid ([here’s the codelist](http://reference.iatistandard.org/203/codelists/ResultType/)), but CoVE says everything is fine. I suspect this is because only embedded codelist values are tested (though I’m unsure why this would be the case?)

Perhaps the message should say: “Your **embedded** codelist values are all valid” ?

andylolz:
This tweaks the copy to avoid the suggestion that non-embedded codelist
values are checked, too.

Fixes #1124.

### DIFF
--- a/cove_iati/templates/cove_iati/explore.html
+++ b/cove_iati/templates/cove_iati/explore.html
@@ -146,9 +146,9 @@
       <div class="panel-body">
         <p class="explanation">
         {% if invalid_embedded_codelist_values %}
-            {% trans "Sorry your data contains values in codelists that do not exist in the standard" %}
+            {% trans "Sorry your data contains values in embedded codelists that do not exist in the standard" %}
         {% else %}
-          {% trans "Congratulations! Your codelist values are all valid" %}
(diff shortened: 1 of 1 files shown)


stevieflow:
Could / should there be a link to http://reference.iatistandard.org/203/codelists-guides/codelist-management/#embedded-codelists ?

andylolz:
Btw I point to CoVE from [IATI Transformer](https://iati-transformer.herokuapp.com) (which does full codelist validation) and a user already contacted me because they were confused about this (IATI Transformer says they have codelist errors, but CoVE says “Congratulations! Your codelist values are all valid”).

andylolz:
Closing in favour of #1139.

Issue #1117, opened 2019-02-14
[1.2] Group validation errors by type


robredpath:
- We already have identifiers for each kind of check
- The work here is to categorise them and present them together
- Risk is that ungrouped errors disappear

robredpath:
https://docs.google.com/document/d/1Jy0kbX7OhP5aY_j67VgTJe9mYcn8iuP3I6ahetfz9OE/edit#bookmark=id.pridbg17tk0 is @mrshll1001 's prototype for this

Bjwebb:
> We already have identifiers for each kind of check

I think I said this. I'm referring to `e.validator` and `e.validator_value`which we have for each error `e` from jsonschema. e.g. for missing fields, `e.validator` is `required`, so we can match for that.

The validation grouping prototype looks good to me.

Bjwebb:
I've split validation errors into 3 groups. Check out the [demo dev deploy](http://1117-group-validation.cove-360-dev.default.threesixtygiving.uk0.bigv.io/?source_url=https://github.com/OpenDataServices/cove/raw/master/cove_360/fixtures/badfile_all_validation_errors.xlsx).

Currently modals are broken. How best to fix that depends on how much more we're changing the tables. The prototype docs only has two columns - is that what we want, or is it just shorthand for the larger table?

PR #1129, opened 2019-02-27
Group validation errors by type

Bjwebb:
#1117


Bjwebb:
This depends on https://github.com/OpenDataServices/lib-cove/pull/9 and https://github.com/OpenDataServices/lib-cove-web/pull/11

BibianaC:
@Bjwebb Travis failed.

Bjwebb:
Looks like its the bulk secret data tests that are failing.

As discussed on standup, I think we can merge this without the bulk secret data tests passing (as they're a bit of a pain to update, and it's mostly as good to do them all in one go).

cove_360/lib/threesixtygiving.py
```diff
@@ -108,10 +109,24 @@ def common_checks_360(context, upload_dir, json_data, schema_obj):
     common_checks = common_checks_context(upload_dir, json_data, schema_obj, schema_name, context)
     cell_source_map = common_checks['cell_source_map']
 
+    validation_errors = context['validation_errors']
+    validation_errors_grouped = defaultdict(list)
```
BibianaC: Would it not be better to extract this to a new function and call it in this one?

cove_360/templates/cove_360/explore.html
```diff
@@ -149,17 +149,65 @@ <h3 class="panel-title panel-title-explore">
...
+              {% include "validation_table.html" %}
+            </div>
+          </div>
+          {% for error_json, values in validation_errors %}
+            {% with error=error_json|json_decode %}
+              {% cove_modal_errors className="validation-errors-"|concat:error_prefix|concat:forloop.counter modalTitle=error.message errorList=values file_type=file_type full_table=True %}
+            {% endwith %}
+          {% endfor %}
+          {% endif %}
+          {% endwith %}
```
BibianaC: Can these three blocks be indented? it is difficult to see where they come from if not. Same for below blocks.

Bjwebb: Agreed, done.

cove_360/lib/threesixtygiving.py
Bjwebb: Yes!

BibianaC:
Thanks for the changes @Bjwebb

Issue #1122, opened 2019-02-14
[1.2] Add the ability to provide richer feedback in additional checks


robredpath:
This is two things:
- Make it so that at least things like bolding, preformatted text and colour work
- (ideally) make it so that we can write copy of the form "x% of your grants have characteristic y" instead of just "x of your grants...."

robredpath:
So, for example, instead of:

**15 grants have a value provided in the Recipient Org: Company Number column that doesn’t look like a company number**
Common causes of this are missing leading digits, typos or incorrect values being entered into this field. 

I'd like to be able to:

**25% of your grants have a value provided in the `Recipient Org: Company Number` column that doesn’t look like a company number**

Common causes of this are missing leading digits, typos or incorrect values being entered into this field. Company numbers are typically 8 digits, possibly starting SC, for example `SC01234569` or `09876543`. You can check company numbers online at [Companies House](https://beta.companieshouse.gov.uk/)

robredpath:
@KDuerden is there any other richness you'd want in Additional Checks stuff?

KDuerden:
@robredpath Thanks - the options you set out above sound good. 
eg bold, colour, hyperlinks and preformatting

PR #1133, opened 2019-02-27
1122 additional checks flex words

BibianaC:
#1122


cove_360/templates/cove_360/explore.html
```diff
@@ -210,7 +210,7 @@ <h3 class="panel-title panel-title-explore">
         {% for message, json_location, spreadsheet_location in quality_accuracy_checks %}
         <tr>
             <td>
-              <strong>{{message.heading}}</strong><br>{{message.message}}
+              <strong>{{ message.heading | safe }}</strong><br>{{ message.message | safe }}
```
Bjwebb: Instead of using the template filter `safe`, I think it's better to use `django.utils.html.mark_safe` in the view ([example](https://github.com/OpenDataServices/cove/blob/28db5ada575aaa81191054ebddfe5962b34f1f39/cove_ocds/lib/exceptions.py#L54)). This is because it brings the declaration of the safety of the string closer to where its being manipulated, which makes it less likely that someone will make it unsafe in future.

BibianaC: Thanks @Bjwebb . I agree with you it is much better to do it with mark_safe.

Issue #1118, opened 2019-02-14
[1.2] Add the ability to have one-from-many additional checks showing their results


robredpath:
eg multiple messages depending on how many of your fundingOrgs have external org IDs

robredpath:
To follow up on our conversation earlier, @BibianaC - the issue that we're trying to address here is that we want to be able to provide more nuanced feedback in the new additional checks sections. 

So, for example, in the check about use of externally referencable org-IDs, we want to be able to say "oh no! less than 30% of your grants use good org-ids, here's some help" OR "30-70% of your grants use decent org IDs - well done, here's some advice on how to make them better" OR ">70% of your grants use decent org IDs, very well done! Your data is already pretty useful. Here's some advice on how to get the final few" OR "100% of your grants use good org-ids, you're a superstar, your data can be used in all these cool ways".

KDuerden:
@robredpath @BibianaC 
the caveat here is 100% is only possible for certain kinds of funders. 50% external org IDs could be excellent for a community foundation that funds small and often unregistered groups. 
So when it comes to the language (and any visual cues) 30% may be framed as 'more is better if possible' rather than a negative.

robredpath:
Thanks @KDuerden - the task here is first of all about putting in place the mechanism to then allow the switching behaviour; but it's helpful to think about the words as well. 

Do you have a sense of how many choices we'll want to present? Would 3 be ok - a negative one, a 'could be better' one, and a 'very good' one? Or, do we want to keep our options open? I'm certainly not thinking of traffic lights or anything like that, after all!

KDuerden:
@robredpath for org IDs I was thinking only 2
x% have external org IDs
0 have external org IDs
I roughed out text/thinking for this in CoVE feedback doc

PR #1134, opened 2019-03-04
1118 one from many msg extra checks

BibianaC:
Set up the ability to add one-from-many additional checks message (issue #1118). 
The new messages are being prepared and will be added in a separate PR.


Bjwebb:
> **Note**: I have checked it manually in my terminal. When I spin a local server I receive an error message: `KeyError: 'validator` about the last work you did @Bjwebb do I need to do anything locally to be able to run it?

You need the most recent commit of lib-cove. Installing from the requirements file should get this.

cove_360/lib/threesixtygiving.py
```diff
@@ -222,8 +224,11 @@ def produce_message(self):
 
     def get_heading_count(self):
         if self.grants_count == 1 and self.count == 1:
+            self.grants_percentage = int(100)
```
Bjwebb: `= 100` should be fine?

cove_360/lib/threesixtygiving.py
```diff
@@ -222,8 +224,11 @@ def produce_message(self):
...
             return '1'
-        return '{:.0%} of'.format(self.count / self.grants_count)
+        heading_percentage = '{:.0%} of'.format(self.count / self.grants_count)
+        self.grants_percentage = int(heading_percentage[:-1])
```
Bjwebb: This doesn't work, because `heading_percentage` contains the string of, so `heading_percentage[:-1]` is e.g. `14% o`.

robredpath:
I really like this approach:
- We can have as many or as few bands for each check as we want
- We can make the boundaries different for each check
- We can easily port the existing checks and improve them as we go

cove_360/lib/threesixtygiving.py
BibianaC: Right! I have fixed that.

cove_360/lib/threesixtygiving.py
BibianaC: Yes

BibianaC:
@robredpath and I have agreed to add the ability to have one-from-many additional checks to all the checks. By doing this, this PR will be able to be merged and be ready when the new validation errors are prepared. (@robredpath @Bjwebb I will let you know when this is the case so you can review it)

@Bjwebb do you know why Flake8 is failing with errors that were not failing before?

Bjwebb:
> @Bjwebb do you know why Flake8 is failing with errors that were not failing before?

When you added `rangedict`, you also updated the other requirements. Looks like the newest version of Flake8 picks up more errors than the older version.

BibianaC:
@robredpath @Bjwebb The PR is more or less ready to review. The `secret_data_tests` is failing in `assert json.load(fp) == json.load(fp_archive)` when checking the difference between both files what I see is that the `fp` has an extra key `validator`. I don't know why this is happening.

Bjwebb:
Sorry, the failing `secret_data_tests` was introduced in this PR https://github.com/OpenDataServices/cove/pull/1129#issuecomment-468210199.

requirements.txt
```diff
@@ -1,41 +1,43 @@
...
+jsonref==0.2
 LEPL==5.1.3
-lxml==4.3.0
-openpyxl==2.5.5
-
-pytz==2018.3
-schema==0.6.7
-six==1.11.0
+lxml==4.3.2
+openpyxl==2.6.1
```
Bjwebb: We're not user if openpyxl 2.6 works with flatten-tool properly - https://github.com/OpenDataServices/flatten-tool/issues/287 (although, it might just be the tests that need updating).

BibianaC: I have changed the version to the original one

BibianaC:
@Bjwebb If you think the code is good then it can be merged with the secret data tests failing.

Issue #1131, opened 2019-02-27
Add spreadsheet-style 'first 3 errors' tables to validation and additional checks


robredpath:

robredpath:
See https://github.com/OpenDataServices/cove/issues/1123#issuecomment-463600797

Bjwebb:
WIP:

![screenshot from 2019-03-06 17-57-41](https://user-images.githubusercontent.com/634/53868509-8c241800-4039-11e9-9c60-70c6e49777a9.png)

Bjwebb:
Here's a demo link - http://1131-spreadsheet-style-validation.cove-360-dev.default.threesixtygiving.uk0.bigv.io/?source_url=https://github.com/OpenDataServices/cove/raw/508dd3c63df3700a7fc264c8e0633536e4f23b8f/cove_360/fixtures/badfile_all_validation_errors_4_times.xlsx
"Missing Fields" doesn't look very good, so scroll down to "Incorrect Formats" to have a look at what it should be like.

![screenshot from 2019-03-07 16-05-02](https://user-images.githubusercontent.com/634/53938262-f5b42d00-40f2-11e9-94fc-04ff908bba28.png)

Clicking on the error count still brings up the old modal. The whole page looks the same for JSON files.

CSVs display a table, but without context atm. (Because I've gone down the, open the spreadsheet in cove route, and have only implemented xlsx so far). [[demo link](http://1131-spreadsheet-style-validation.cove-360-dev.default.threesixtygiving.uk0.bigv.io/?source_url=https://github.com/OpenDataServices/cove/raw/508dd3c63df3700a7fc264c8e0633536e4f23b8f/cove_360/fixtures/fundingproviders_grants_2_grants.csv)]

![screenshot from 2019-03-07 16-12-42](https://user-images.githubusercontent.com/634/53938522-df5aa100-40f3-11e9-8206-8f2f2d1c2f00.png)

"Missing Fields" does not look good because there's no column letters. We could revert back to the old view, but that would be inconsistent, or work out how to make a better table.

![screenshot from 2019-03-07 16-08-17](https://user-images.githubusercontent.com/634/53938351-517eb600-40f3-11e9-9af0-3368782004fd.png)

Bjwebb:
Here's a suggestion for displaying the missing fields:
[...]

PR #1137, opened 2019-03-07
Spreadsheet style validation error tables

Bjwebb:
#1131 

[Screenshots and demo links.](https://github.com/OpenDataServices/cove/issues/1131#issuecomment-471427123)

In order to provide context cells for the tables, I'm re-opening the xlsx file from within CoVE (previously just within flatten-tool).

All of these changes are specific to 360Giving CoVE only. As part of this I've made a copy of `validation_table.html`.


cove_360/lib/threesixtygiving.py
```diff
@@ -104,31 +106,115 @@ def get_grants_aggregates(json_data):
...
+        prev_number = next(prev_numbers, -1)
+        next_number = next(next_numbers, None)
+        if prev_number != number - 1 and prev_number + 1 != number - 1 and number - 1 > 0:
+            yield number - 1
+        yield number
+        if next_number != number + 1:
+            yield number + 1
+
+
+def spreadsheet_style_errors_table(examples, openpyxl_workbook):
```
BibianaC: Is it possible to add a test?

BibianaC:
@Bjwebb I guess it is fine what you wrote but in this case I would feel more comfortable if someone else reviews it too.

cove_360/lib/threesixtygiving.py
Bjwebb: Good idea, I've done so.

Bjwebb:
@BibianaC Thanks. I got some feedback from Rob on a call, so I'll be working through that, and then I'll assign it to someone else to review the code.

PR #1138, opened 2019-03-13
Fix additional checks URL

BibianaC:
On my first PR I introduced two templates about the extra checks that in reality were not used. With this PR I wanted to fix that. 
It is about displaying additional checks messages. This url https://dataquality.threesixtygiving.org/additional_checks

I had to move the `mark_safe` in headings inside the `check_text` because is in here where the code takes the information.
We do have now a problem with the message as the way it is done with the `RangeDict()` it doesn't display anything. I have been trying different things without success. Do you have an idea of how to fix this?

In this PR appear changes that should already be in master-360-data-quality-tool.


Bjwebb:
Could you rebase this onto `master-360-data-quality-tool `?

BibianaC:
@Bjwebb I did this branch from the master-360 branch. Yes, I will rebase and see if it is fixed. I also need to fix the merge conflicts.
The problem about messages not being displayed will still be there.

Bjwebb:
Thanks. Yes, I'm not sure what best to do with the messages. What we really want is for `RangeDict` to support `.items()` or `.values()` properly (which it doesn't). One approach could be to create an OrderedDict with the same key (which can be iterated over from the "additional_checks" page), and then use that to create the RangeDict (possibly within `AdditionalTest.__init__).

BibianaC:
@Bjwebb @robredpath files are now correct. Many thanks for your help @Bjwebb !
The objective of this PR was to show all the test classes in the url additional_checks. While doing this, I realise that the message is not longer displaying because of the RangeDict. I think this PR can be reviewed as it is and I will create a separate branch to fix the message problem.

PR #1142, opened 2019-03-14
Additional checks url display msg

BibianaC:
Additional checks URL was not displaying messages. In this PR, the RangeDict library is override to use it as a dict and fix the problem.


cove_360/templates/cove_360/additional_checks.html
```diff
@@ -15,7 +15,10 @@ <h2> Checks </h2>
     <p>
       <label>Class name</label> {{check.class_name}}<br/>
       <label>Heading</label> {{check.heading}}<br/>
-      <label>Message</label> {{check.message}}<br/>
+      <label>Message</label>
+      {% for percentage_range, message in check.message %}
+         {{message}}<br/>
```
robredpath: @BibianaC were you planning on displaying the % range here?

BibianaC: @robredpath not now as there is only one message. Once we have more than once then yes.

PR #1148, opened 2019-03-25
Specialist language

robredpath:


cove_360/templates/cove_360/explore.html
```diff
@@ -219,9 +238,9 @@ <h4>Other</h4>
     <div class="panel-heading" data-toggle="collapse" data-target="#quality-accuracy-body">
       <h3 class="panel-title panel-title-explore">
         <span class="font-tick question" aria-hidden="true"></span>
-          {% trans "Quality / Accuracy" %}
```
BibianaC: I thought we were going to leave `accuracy` in the check name.

BibianaC:
@robredpath some tests are failing.

cove_360/templates/cove_360/explore.html
robredpath: @KDuerden 's feedback was that she wanted a single-word title, and I think I agree. I've also kept the internal naming of quality-accuracy as we might want to change that in the future!

robredpath:
@BibianaC tests now pass :)

PR #1161, opened 2019-04-08
[1158] Non Embedded codelist check

kindly:
Add script to fetch codelists.
Refactor process_codelists a bit to make path the key instead of name.


odscjames:
One alternative to putting these in git is to put the folder in .gitignore, and change the instruction to "run this script before runnning IATI cove" and then put the command in the salt so it's run on each deploy, and get salt to put it in cron so it's run every weekend.

Was that thought of and rejected?

kindly:
@odscjames  yes that is an option and possibly something we would want to do at some point, with probably a fallback.

I did consider it but:
- I did not want to add the additional step for anyone using the cli tool and tests.
- It would be good to make sure the tests are definitely reproducible
- The cadence of change of the codelists is slow.

However I also do not very much like it as it is either.  I just considered it the least worse (and easiest) option.

cove_iati/lib/process_codelists.py
```diff
@@ -102,7 +122,7 @@ def is_int(value):
 
 
 def invalid_codelist_values(codelist_values, filename, source_map=None):
-    path_to_codelist = create_path_to_codelists(codelist_values)
+    #path_to_codelist = create_path_to_codelists(codelist_values)
```
odscjames: Is this meant to be removed totally, or should it be commented back in?

I'm not sure we have a normal practice of leaving commented code.

odscjames:
Out of curiosity, how does `{% block non_embedded_codelists %}` work? So the blocks contents are defined here, but shouldn't there be a block with the same name somewhere else for that content to appear in? But I can't find that .... but it works, so something is happening ...

odscjames:
The rest of the code seems fine as far as I can follow, but having not seen this code or IATI before I do find it hard to follow.

odscjames:
> I did not want to add the additional step for anyone using the cli tool and tests.

We already have like 17 steps, an extra one isn't going to hurt :-P We could even have the code check it and do itself if needed.

> It would be good to make sure the tests are definitely reproducible
> The cadence of change of the codelists is slow.

Those are fair points tho - ok.

cove_iati/lib/process_codelists.py
kindly: yes that should be removed

kindly:
> Out of curiosity, how does `{% block non_embedded_codelists %}` work? So the blocks contents are defined here, but shouldn't there be a block with the same name somewhere else for that content to appear in? But I can't find that .... but it works, so something is happening ...

I just am using block as a way to document the sections without needing an explicit comment with the very unlikely possibility that that section could be overridden at some point.

cove_iati/lib/process_codelists.py
kindly: It has now

PR #1163, opened 2019-04-09
Fix padding in validation error groups

robredpath:


cove_360/templates/cove_360/explore.html
```diff
@@ -179,7 +179,7 @@ <h3 class="panel-title panel-title-explore">
                   <h4>Missing Fields</h4>
                 </div>
                 <div>
-                  <p>Some or all of your entries were missing fields which are required by the 360Giving Standard.</p>
+			<p class="panel-body">{% trans "Some or all of your entries were missing fields which are required by the 360Giving Standard." %}</p>
```
BibianaC: Do we need translation in 360?

PR #1165, opened 2019-04-17
Update requirements with v0.4.0 of lib-cove-ocds

BibianaC:


odscjames:
We also need to upgrade  lib-cove to v0.6.0 at the same time - there is a note about that in https://github.com/open-contracting/lib-cove-ocds/blob/master/CHANGELOG.md but we should find a way to be clearer about this in future!

BibianaC:
Duplicated with branch update-libs

Issue #901, opened 2017-11-28
Additional Checks for OCDS


robredpath:
The OCDS Validator already does more than simple validation. We might want to implement additional checks similar to those in 360's DQ tool. Whether or not we do depends on a few things:
- Whether we want to 'keep it simple' 
- Whether we believe that displaying this kind of information in the Validator will actually cause a change in behaviour / improvement of the data that publishers actually publish
- Whether there are better ways (value for money?) than the Validator for stimulating this change.

jpmckinney:
Today's call considered some additional checks, which are along the lines of keeping it simple, while adding some high priority checks that we believe will produce an improvement of the data.

So, we might close this issue and open more specific ones.

robredpath:
Added to the April sprint. 

From the sprint doc: 



> Add an extensible mechanism to the DRT to carry out inspection of supplied data in order to give helpful feedback on issues with data that may suggest an issue but that do not cause data to be invalid against schema.
> 
> Use the above mechanism to test for at least one, and ideally both, of:
> - Checking that buyer and suppliers are listed in the parties array of 1.1 data (with the appropriate roles)
> - Identifying when fields, objects and arrays exist but are empty or contain only whitespace
> 
> 
> 

```

robredpath:
Note that for: 

> Checking that buyer and suppliers are listed in the parties array of 1.1 data (with the appropriate roles)

We need to check parties arrays across all items with the same ocid 

We don't need to compile the release to see if parties get deleted.

robredpath:
> Identifying when fields, objects and arrays exist but are empty or contain only whitespace 

-: is done for releases. Still to do - making it work for packages. 

We didn't get to the parties array piece.

jpmckinney:
Regarding the next sprint (and whether to include the parties check in CoVE): non-schema checks will also be implemented in the [data quality tools](https://open-contracting-partnership.forms.fm/ocds-data-quality-tools/forms/6481) project, so we'll need to decide which additional checks we want to alert publishers about during their implementation (i.e. in CoVE) – versus which we want to follow-up about when checking the entire dataset (e.g. some checks are less likely to fail on the samples that are input into CoVE). I think the parties check is still a good one for CoVE, but we can double-check.

robredpath:
[...]

PR #1170, opened 2019-04-26
901 add additional checks

BibianaC:
Relates to #901


BibianaC:
@kindly I suppose that once you have reviewed it and before merging it, I will have to do something for translations?

duncandewhurst:
Tested locally and this looks good, only suggestion is to make the 'Additional checks' box the same colour (yellow) as the 'Additional codelist values' box, since they aren't actually validation errors:

image

Happy for this to be merged otherwise

Issue #1208, opened 2019-08-17
[IATI] Og file XML > spreadsheet output


stevieflow:
This is mentioned in #959 and also via https://github.com/OpenDataServices/flatten-tool/issues/151 - but I wanted to log a specific issue in terms of the input of an Org XML file into CoVE does result in schema and ruleset tests - and then offers a conversion to spreasheet

As documented in #959 - this generates a blank spreadsheet

It would be good to fix this

Bjwebb:
@stevieflow Here's a dev deploy — http://1208-iati-cove-orgxml-spreadsheet.dev.cove.opendataservices.coop/ — could you check that's behaving as expected?

robredpath:
@stevieflow When you get a sec, could you check this so that we can close this issue if it's working as expected?

Bjwebb:
I think we got the feedback we needed from @stevieflow on the [pull request](https://github.com/OpenDataServices/cove/pull/1219).

stevieflow:
Yeah, thanks @Bjwebb - I think we have that in the PR @robredpath (thanks for checking though)

PR #1219, opened 2019-08-30
#1208 iati cove orgxml spreadsheet

Lathrisk:
Pass through the relevant `root_list_path` to libcove when organisations are, or are not, identified in the uploaded XML. 

Part of https://github.com/OpenDataServices/cove/issues/1208


Lathrisk:
@Bjwebb requirements files are now updated (pyenv was mismanaging python, I suspect as a consequence of my mismanagement of pyenv).

Bjwebb:
Here's a dev deploy: http://1208-iati-cove-orgxml-spreadsheet.dev.cove.opendataservices.coop/

@Lathrisk FYI here's the change I had to make in the deploy repo: https://github.com/OpenDataServices/opendataservices-deploy/commit/5f4ba3c8d26f1ddbd2e570f2504d650a5145eda9

stevieflow:
Thanks @Lathrisk @Bjwebb & @rory09 for review.

Yes, agree that the missing identifier will need addressing

Also - I'm not too sure if we also need to output a Meta tab with this too - as there is the control within this to force CoVE to convert to the org schema...?

https://docs.google.com/spreadsheets/d/1MXjepDgfzKw0ULUWzKmXVS0Al6rMhMHlgN-ZF6hLc70/edit#gid=91665843

Lathrisk:
Thanks @stevieflow and @rory09 for the reviews. I suspect the scope of this piece of work might creep and we should perhaps raise these as separate issues instead. I suspect the fact that CoVE now outputs a spreadsheet covers the scope of this ticket, but: 

If you feel that the missing information in the output is worse than nothing at all then we shouldn't merge this until those issues are resolved.

Otherwise I would suggest that we progress this work and raise those issues (missing `organisation-identifier` and `Meta` tab) separately.

@Bjwebb, perhaps you can advise? Are the issues raised by @stevieflow and @rory09 issues related to the flatten tool itself, rather than CoVE?

Bjwebb:
I think fixing the `organisation-identifier` issue is important to do before this is merged. It's really quite fundamental to being able to use the spreadsheet, and it can be fixed by passing another argument to flatten-tool (`root_id`). 

As for the Meta tab, it's true that we can't reupload the spreadsheet to CoVE without it. However, this will require work on flatten-tool to add a Meta tab when flattening. There's an issue for this (https://github.com/OpenDataServices/flatten-tool/issues/176) but it doesn't have much info yet. Since this is a larger piece of work, I would treat it as a separate issue if possible, and manually add the Meta tab as needed for a bit.

Lathrisk:
@stevieflow @rory09 

I've attached the output from a new flattened version of the example file (referenced by Rory - https://github.com/OpenDataServices/cove/pull/1219#pullrequestreview-284115831).

This now includes the organisation-identifier in the left most column.
[flatten-example.xlsx](https://github.com/OpenDataServices/cove/files/3590920/flatten-example.xlsx)

stevieflow:
Thanks @Lathrisk 

I can see the org-identifier in the output file, so that's good

This does throw up another (albeit indirect & interesting) question in that in the output file we have organisations and documents that are in multiple languages / multiple categories, which are split over different sheets.  Should we (for example) have two documents in two categories, but under the same organisation (highly likely) then there'd be no way to stitch this back together (partly because of the lack of UID in the IATI standard for nested elements, but also as a consequence of this behaviour of CoVE

Again, I dont see this as a blocker, but it could render the data rather useless.  I wonder if there's something we can do here though ... --> @Bjwebb @rory09 ?

Bjwebb:
@stevieflow  Yeah, its not an easy one to fix. I think creating our own IDs that get ignored or namespaced by unflatten is probably the way to defal with this. The more difficult alternative is to change the shape of data flatten-tool produces (more columns instead of extra sheets).

PR #1223, opened 2019-09-10
Add CSV export to 360Giving Additional Checks report

robredpath:


robredpath:
@kindly @michaelwood this is to add a CSV export to the additional checks page so that we can use a spreadsheet to communicate with 360 about changes to the additional checks

Bjwebb:
If the CSV might contain unicode, it could be good to pop a BOM at the start, so Excel will open it correctly. (Like we [do for Grantnav](https://github.com/ThreeSixtyGiving/grantnav/blob/26e3cb71aab1b20a62d34f87899550c0dd202f9a/grantnav/frontend/views.py#L144)).

robredpath:
Good point, thanks @Bjwebb . It's unlikely to use UTF-8, right up until we try and it breaks.

PR #1247, opened 2019-11-13
Use libs from Pypi instead of github

rhiaro:


odscjames:
Basically some comment as https://github.com/OpenDataServices/lib-cove-web/pull/32#issuecomment-553485192

rhiaro:
PR out of date

PR #1256, opened 2020-01-09
Remove OCDS code and docs

rhiaro:
In favour of github.com/open-contracting/cove-ocds


Bjwebb:
Tests are broken due to `hypothesis`, as with https://github.com/OpenDataServices/cove/pull/1257#issuecomment-572601162

rhiaro:
[Refactor OCDS-specific code into a new application](https://trello.com/c/6SyOfmT6/49-refactor-ocds-specific-code-into-a-new-application)

odscjames:
The live server has now been changed over to the new repository. If no problems reported in a bit, this can be rebased, fixed up and actioned.

docs/deployment.md
```diff
@@ -4,25 +4,10 @@ General Django deployment considerations apply to deploying Cove. We deploy usin
 
 ## Before a live deploy
 
-Travis tests will fail if a branch isn't ready to be merged and deployed. This includes if OCDS translations are missing.
+Travis tests will fail if a branch isn't ready to be merged and deployed. This may happen if translations are missing.
```
odscjames: "This may happen if translations are missing." I don't think we do any translation tests any more (360 and IATI don't have any) so I think this line should be removed

odscjames:
I've had a look, this looks good.  See one comment above.

Can we remove

```
  # Install translate toolkit to get pocount binary
  - sudo apt-get install translate-toolkit
```

from Travis too?

    libcoveocds>=0.7.4


Can be removed from requirements.in 

It will need rebasing tho, and when it does it will show a change to requirements.* that is just a comments change that came up because the same fix was applied in 2 different branches. It can be safely dropped.

I don't know when we will be merging this in tho, so I haven't started any rebasing work just yet.

docs/deployment.md
Bjwebb: That's correct.

Bjwebb:
Yes, remove `translate-toolkit` from Travis, and remove `libcoveocds` from the requirements.

odscjames:
I tried to rebase but it got complicated so I just start again.

odscjames:
This does everything I know of apart from remove libcoveocds and rebuild reqs. Travis is currently having a cache problem and refuses to install when I try that. So let's just get this in now, then we'll try the last part of doing the reqs in a seperate pull request (Also, we don't have to deploy 360 and IATI after this pull request and we probably should after changing reqs)

PR #1257, opened 2020-01-09
Update libcove, and other requirements

rhiaro:
This libcove update fixes the large blocks of JSON error reporting and oneOf validation message issues.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,10 +1,10 @@
 pip
 Django<1.12
 #^^ rq.filter: <1.12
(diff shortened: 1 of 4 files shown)


Bjwebb:
Looks like `hypothesis` is broken.

rhiaro:
It was coverage that broke things; need to make travis run it in a different way (https://github.com/nedbat/coveragepy/issues/862#issuecomment-572318505) but for now pinned the version

PR #1273, opened 2020-03-12
Mw/upgrade lib cove web 0.12.0

michaelwood:


requirements.txt
```diff
@@ -1,7 +1,7 @@
 Django==1.11.29 # rq.filter: <1.12
 flattentool==0.11.0
 libcove==0.15.0
--e git+https://github.com/OpenDataServices/lib-cove-web.git@55dad570493c9429a4821760c599b4a84f099304#egg=libcoveweb
+-e git+https://github.com/OpenDataServices/lib-cove-web.git@36ad4df14916d7be28b2c1332427ad5140b35ed5#egg=libcoveweb
```
BibianaC: Should this not be `@v0.12.0` instead of the commit?

michaelwood: Yes I agree ideally, but update_requirements doesn't do that. I could manually overwrite the output but this hasn't been done in the past.  I would be up for us doing that or fixing update_requirements but don't want to make it inconsistent with our approach elsewhere.

PR #1274, opened 2020-03-23
Update version lib cove 0.16.1

BibianaC:


michaelwood:
Looks good, can you fixup/squash the duplicated commits together so that you only have 2 commits.

BibianaC:
@michaelwood I have squashed the last two commits in one. The other ones I prefer them to be separated.

BibianaC:
@michaelwood not sure, I have updated requirements and this is the result of it.

michaelwood:
> @michaelwood not sure, I have updated requirements and this is the result of it.

There is no way to easily isolate the difference between running update_requirements after bumping libcove version and generally running update_requirements. I think that is actually why I said it would be best to do it as two commits.

Issue #1200, opened 2019-07-23
Quality feedback messages in 360Giving Quality Checks should give number of grants, not %, if it's a small number


robredpath:
From @KDuerden

>  Quality feedback messages to give number of grants instead of percentage
> The messages for quality section are specific and framed as things to be looked at and potentially fixed, so a count makes more sense in this context. These tend to be feedback where there are few examples in large files means it is possible to have a 0% of x are y message 

<img width="859" alt="Zero percent Quality check" src="https://user-images.githubusercontent.com/641009/61729118-7b690a00-ad6e-11e9-9b93-69adf62d6ef4.PNG">

robredpath:
For large data sets, it's much more informative to say "95% of your grants...."

When it's a very small number, 0% is misleading. We should just say the number. 

@michaelwood this could be a good candidate for you to look at as part of your iterative work, if it appeals!

KDuerden:
On this one, I'm of the view that messages in the Quality section should always be numbers rather than percentages. This is because the expectation is these are issues to be fixed, so being specific makes sense to me. I don't agree that using a percentage is more effective on large data sets/one with high proportion of errors. Using a number can still convey the "something isn't right here" message effectively. 
In the Usefulness section I do think there's an arguement for % unless the result is eg 5% or below.

BibianaC:
@KDuerden Is it ok for you if I work on this one?

KDuerden:
Thanks @BibianaC that would be great!

BibianaC:
![](https://github.trello.services/images/mini-trello-icon.png) [Quality feedback messages in 360Giving Quality Checks should give number of grants, not %, if it's a small number](https://trello.com/c/ttCKtbdv/43-quality-feedback-messages-in-360giving-quality-checks-should-give-number-of-grants-not-if-its-a-small-number)

PR #1277, opened 2020-04-02
Additional checks amount output

BibianaC:
relates to #1200 

- Quality section should always be numbers.
- Usefulness section % unless the result is below 5%, in which case it switches to numbers.


cove_360/lib/threesixtygiving.py
```diff
@@ -285,13 +285,14 @@ def common_checks_360(context, upload_dir, json_data, schema_obj):
...
+            json_data, cell_source_map, TEST_CLASSES[test_class_type], ignore_errors=True, return_on_error=None)
+
         context.update({
-            '{}_errored'.format(test_classes_type): extra_checks is None,
-            '{}_checks'.format(test_classes_type): extra_checks,
-            '{}_checks_count'.format(test_classes_type): (len(extra_checks) if extra_checks else 0) + (1 if context['data_only'] else 0)
+            '{}_errored'.format(test_class_type): extra_checks is None,
+            '{}_checks'.format(test_class_type): extra_checks,
+            '{}_checks_count'.format(test_class_type): (len(extra_checks) if extra_checks else 0) + (1 if context['data_only'] else 0)
         })
```
michaelwood, line 297: Is this related to "quality checks in numbers" ? if not it is a good refactor that could be in a separate commit.

cove_360/lib/threesixtygiving.py
```diff
@@ -911,7 +918,7 @@ def process(self, grant, path_prefix):
                         self.json_locations.append(id_location)
                         self.count += 1
 
-        self.heading = self.format_heading_count(self.check_text['heading'])
+        self.heading = self.format_heading_count(self.check_text['heading'], test_class_type='quality_accuracy')
         self.message = self.check_text['message'][self.grants_percentage]
```
michaelwood, line 926: A "nice to have" would if we could start using a const for this value instead of typing it out each time.

cove_360/lib/threesixtygiving.py
```diff
@@ -997,6 +1004,7 @@ def process(self, grant, path_prefix):
 def run_extra_checks(json_data, cell_source_map, test_classes):
     if 'grants' not in json_data:
         return []
+
     test_instances = [test_cls(grants=json_data['grants']) for test_cls in test_classes]
```
michaelwood, line 1011: extra line?

cove_360/lib/threesixtygiving.py
BibianaC: Really good comment, thanks! I didn't like having so much repetition but it didn't come to do it as simple as adding a constant.

cove_360/lib/threesixtygiving.py
BibianaC: I have deleted it.

cove_360/lib/threesixtygiving.py
BibianaC: You are right, this should had been in a different commit. I will leave it as it is now if you don't mind.

BibianaC:
[Quality feedback messages in 360Giving Quality Checks should give number of grants, not %, if it's a small number](https://trello.com/c/ttCKtbdv/43-quality-feedback-messages-in-360giving-quality-checks-should-give-number-of-grants-not-if-its-a-small-number)

Issue #1301, opened 2020-10-02
IATI - refresh of codelist values needed?


stevieflow:
@Bjwebb @kindly @BibianaC  I think our store of IATI codelists might be out of date/synch - there's Non Embedded codelist value errors, when the codes are valid (@siwhitehouse has an example)

This might be a better source of data from which to pull the codelist - https://codelists.codeforiati.org/ - but I'm not sure what the underlying process currently is...

Bjwebb:
We have [a command to keep these in sync](https://github.com/OpenDataServices/cove/blob/master/get_iati_non_embedded_codelists.py), but I don't think its run automatically, meaning they're out of date.

stevieflow:
Thanks

Please can we run it , to get back into synch, in the meantime?

kindly:
This is merged and deployed.  @stevieflow could you check to see if the results (code-list values) are what you expect.

stevieflow:
Thanks @kindly 

@siwhitehouse want to check this via the SDG vocab you observed?

stevieflow:
I tested some data with an SDG vocabulary - and this pass without a fail warning

stevieflow:
> We have a command to keep these in sync, but I don't think its run automatically

@Bjwebb @kindly @BibianaC  to avoid this (see #1354 for latest example), could the command be autorun with a github command?

(suggestion ^ from @andylolz , via code4IATI channels)

andylolz:
Thanks @stevieflow,

Specifically on the [Reporting Organisation codelist](https://codelists.codeforiati.org/ReportingOrganisation/) – this codelist seems to update pretty frequently (more than once per week). You can get a sense of that from [its git history](https://github.com/codeforIATI/Unofficial-Codelists/commits/master/xml/ReportingOrganisation.xml). So auto-checking for updates pretty frequently (weekly or even daily) would help prevent validation false negatives.

stevieflow:
Thanks @andylolz 

Yes - with the new org references check, being in synch with the Reporting Org codelist will have to be automated, to save multiple requests to @kindly @BibianaC and/or @Bjwebb !

stevieflow:
On a backlog for @odscjames !

Bjwebb:
FWIW, the new org ref work pulls from codeforiati each time https://github.com/OpenDataServices/cove/blob/live/cove_iati/lib/iati.py#L499
It would probably be good to be more consistent about this though.

ghost:
[...]

PR #1302, opened 2020-10-08
iati cove: update codelists

kindly:
https://github.com/OpenDataServices/cove/issues/1301


Bjwebb (approved):
I've not reviewed the changes in depth, but looks like everything's been pulled correctly.

Bjwebb (approved):
Doesn't look like anything's changed, so re-approving. I think you requested my review after I'd already given it, which invalidated it!

PR #1312, opened 2020-10-23
Changes for compatibility with latest libcove/libcoveweb

Bjwebb:


Bjwebb:
@KDuerden @morchickit  @drkane This PR will bring the 360Giving Data Quality Tool up to date with changes in our shared libraries lib-cove and lib-cove-web (which are also used for OCDS, OC4IDS, IATI and BODS). This will have a couple of expected changes, and as with any software update might have changes we didn't intend, so do get in touch if you notice anything odd.

Expected changes:
* Add the word 'Co-operative' to our company name under 'built by'.
* Limit the number of error locations shown when clicking on Error Count to 1000. (See screenshot below). This is to reduce the size of the html page, and make it quicker to load.

Screenshot from 2020-10-28 16-27-15

BibianaC (approved):

PR #1314, opened 2020-11-04
Travis -> GitHub Actions

ghost:


odscjames:
I did try splitting the test file to test-360 and test-iati but then I wasn't sure how Coveralls would work so I left it for now. If we get the basic changeover in, details like that could be looked at later?

odscjames:
[Switch from Travis to GitHub Actions](https://trello.com/c/sLVHLfoo/46-switch-from-travis-to-github-actions)

Bjwebb (approved):

PR #1326, opened 2021-02-12
cove360: additional checks: dates

BibianaC:
start date before end date, impossible dates, far future and far past dates are flagged up.

### DIFF
--- a/cove_360/lib/threesixtygiving.py
+++ b/cove_360/lib/threesixtygiving.py
@@ -1,4 +1,5 @@
 import datetime
+import functools
 import itertools
 import json
 import re
@@ -16,7 +17,6 @@ from rangedict import RangeDict as range_dict
 QUALITY_TEST_CLASS = 'quality_accuracy'
 USEFULNESS_TEST_CLASS = 'usefulness'
 
-GRANT_DATES = {}
 DATES_JSON_LOCATION = {
     'award_date': '/id',
     'planned_start_date': '/plannedDates/0/startDate',
@@ -1006,8 +1006,7 @@ class ImpossibleDates(AdditionalTest):
     )
 
     def process(self, grant, path_prefix):
-        id = grant.get('id')
-        grant_dates = GRANT_DATES.get(id)
+        grant_dates = create_grant_dates_dict(Grant(grant))
 
         if grant_dates:
             for date_type, date_format_error in (
@@ -1047,8 +1046,7 @@ class PlannedStartDateBeforeEndDate(AdditionalTest):
     )
 
     def process(self, grant, path_prefix):
-        id = grant.get('id')
-        grant_dates = GRANT_DATES.get(id)
+        grant_dates = create_grant_dates_dict(Grant(grant))
 
         if grant_dates:
             planned_start_date = grant_dates.get('planned_start_date', {}).get('datetime_date')
@@ -1079,8 +1077,7 @@ class ActualStartDateBeforeEndDate(AdditionalTest):
     )
 
     def process(self, grant, path_prefix):
-        id = grant.get('id')
-        grant_dates = GRANT_DATES.get(id)
+        grant_dates = create_grant_dates_dict(Grant(grant))
 
         if grant_dates:
             actual_start_date = grant_dates.get('actual_start_date', {}).get('datetime_date')
@@ -1110,8 +1107,7 @@ class FarFuturePlannedDates(AdditionalTest):
     )
 
     def process(self, grant, path_prefix):
-        id = grant.get('id')
-        grant_dates = GRANT_DATES.get(id)
+        grant_dates = create_grant_dates_dict(Grant(grant))
 
         if grant_dates:
             for date_type, input_date in (
@@ -1143,8 +1139,7 @@ class FarFutureActualDates(AdditionalTest):
     )
 
     def process(self, grant, path_prefix):
-        id = grant.get('id')
-        grant_dates = GRANT_DATES.get(id)
+        grant_dates = create_grant_dates_dict(Grant(grant))
 
         if grant_dates:
             for date_type, input_date in (
@@ -1175,8 +1170,7 @@ class FarPastDates(AdditionalTest):
     )
 
     def process(self, grant, path_prefix):
-        id = grant.get('id')
-        grant_dates = GRANT_DATES.get(id)
+        grant_dates = create_grant_dates_dict(Grant(grant))
 
         if grant_dates:
             for date_type, input_date in (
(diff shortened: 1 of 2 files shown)


cove_360/lib/threesixtygiving.py
```diff
@@ -979,6 +990,180 @@ def process(self, grant, path_prefix):
         self.message = mark_safe(self.check_text['message'][self.grants_percentage])
 
 
+class ImpossibleDates(AdditionalTest):
+    """Check if dates supplied are plausible (eg no 31st Feb)"""
```
robredpath: Let's also say here that we check for dates that are plausible but didn't happen (eg 29th of Feb in a non-leap year) because this message is used by us and Katherine to remember what these checks do.

cove_360/lib/threesixtygiving.py
```diff
@@ -979,6 +990,180 @@ def process(self, grant, path_prefix):
...
+
+    check_text = {
+        "heading": mark_safe("dates that don't (or won't) exist"),
```
robredpath: I'm no longer happy with this wording. 

Let's change it to "dates that didn't, or won't, exist".

cove_360/lib/threesixtygiving.py
```diff
@@ -1,19 +1,30 @@
...
+import pytz
+from dateutil.relativedelta import relativedelta
 from django.utils.html import mark_safe
 from libcove.lib.common import common_checks_context, get_orgids_prefixes
 from rangedict import RangeDict as range_dict
 
 QUALITY_TEST_CLASS = 'quality_accuracy'
 USEFULNESS_TEST_CLASS = 'usefulness'
 
+GRANT_DATES = {}
```
robredpath: This mechanism is a great idea that will help to speed up the runtime of this function. This implementation doesn't feel super Pythonic, though. I've done a quick bit of Googling and it looks like there's a range of memoization (I learned a new word today!) decorators that achieve the same ends but are more Pythonic and I think also make this code a bit less fragile. 

Cc-ing some actual Python devs (@kindly ? @Bjwebb ?) for their view, though!

cove_360/lib/threesixtygiving.py
```diff
@@ -989,6 +1174,11 @@ def process(self, grant, path_prefix):
         OrganizationIdLooksInvalid,
         MoreThanOneFundingOrg,
         LooksLikeEmail,
+        ImpossibleDates,
```
robredpath, line 1233: While I love this file as an interview question, if you did fancy refactoring it into a directory full of includes (and maybe even some fancy autofunc stuff, if Python lets you do that?) then I wouldn't object. But no worries if you don't fancy doing that right now.

cove_360/tests.py
```diff
@@ -22,7 +25,7 @@
                 'Sponsor(s)': ' ',
                 'Surname of applicant': 'Roe',
                 'amountAwarded': 0,
-                'awardDate': '24/07/2014',
+                'awardDate': '{}-07-24'.format(current_year),
```
robredpath: There's probably a reason why this date was formatted using `DD/MM/YYYY`, so I'm wary of us changing it. @Bjwebb any ideas?

robredpath (changes requested):
Looking good, just a few things to sort out before we can get this merged :)

cove_360/tests.py
Bjwebb: Not sure, it could just be bad data copy/pasted here. Since this data is only used to test additional checks, I think the tests would fail if they relied on this.

cove_360/lib/threesixtygiving.py
Bjwebb: Yeah, I'm generally in favour of memoisation decorators, although I've not had chance to review what this does exactly.

BibianaC: I didn't know about `memoisation decorators`. I will look into it. Thanks @robredpath!

@Bjwebb there are a few checks for dates. Instead of going through the dates and format them in each check, it is done once at the beginning and store the format dates and error messages in this dictionary. Then in each check the date or error message is check directly in the dict.

cove_360/lib/threesixtygiving.py
BibianaC: I will change it to 

Check if dates supplied are plausible (eg no 31st Feb) or are plausible but didn't happen (eg 29th of Feb in a non-leap year).

cove_360/lib/threesixtygiving.py
BibianaC: I have updated the wording.

cove_360/lib/threesixtygiving.py
BibianaC: @robredpath @Bjwebb I have been looking at memoisation and I think it might not make the function much more efficient but correct me if I'm wrong.

BibianaC:
[Additional Check - do dates make sense?](https://trello.com/c/Qa3XTG6g/97-additional-check-do-dates-make-sense)

cove_360/lib/threesixtygiving.py
robredpath: Sounds good!

cove_360/lib/threesixtygiving.py
robredpath: @BibianaC as I understand it, you've basically re-implemented memoization here, but in a way that's a bit more fragile. I don't think that 'real' memoization would actually be any faster, because it would be doing basically the same thing. 

The benefit of doing it using a memoization decorator would be that the code would be a bit easier to read (global-scope dicts are fairly uncommon in Python, so it's a bit jarring to see) and it would make the code easier to refactor in the future, if we wanted to. 

It might also save on some memory - for a very large file, the dict could get pretty big. It looks like [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) lets you limit the size of the cache. 

To be clear - I'm suggesting:
* Refactoring `create_grant_dates_dict` to operate on a single grant each time 
* Calling that function instead of accessing the dict in each of the additional checks that uses the data

That said, this isn't a blocker for me; just a suggestion! I'd be perfectly happy to keep this as it is, and to make a note of a potential improvement for the future.

BibianaC: @robredpath I have implemented it without a decorator though. 
I have disabled the max size of the cache but I can modify it. The default one is 128.

michaelwood:
When this is ready it will need to be patched on to https://github.com/ThreeSixtyGiving/dataquality/blob/dev/lib360dataquality/cove/threesixtygiving.py as well, which I can do.

robredpath (approved):
I've not tried to run the code, but the changes look like what I've asked for!

PR #1343, opened 2021-05-12
Re-add dealer and sentry-sdk, as they're gone from libcoveweb

Bjwebb:
Same PR for cove-bods: https://github.com/openownership/cove-bods/pull/61


cove_iati/settings.py
```diff
@@ -9,8 +9,26 @@
...
 )
 
+# We use the setting to choose whether to show the section about Sentry in the
+# terms and conditions
+SENTRY_DSN = env('SENTRY_DSN')
+
+if SENTRY_DSN:
+    import sentry_sdk
+    from sentry_sdk.integrations.django import DjangoIntegration
+    from sentry_sdk.integrations.logging import ignore_logger
```
BibianaC, line 22: The imports can't be at the beginning of the file?

BibianaC (approved):

cove_iati/settings.py
Bjwebb: I'm not sure, this is how it was in lib-cove-web.

BibianaC: It is not a blocker. I have approved the PR but I was wondering if it could go at the beginning. Leave them as they are then. Thanks.

PR #1352, opened 2021-06-29
requirements: Update all requirements

ghost:


ghost:
This bumps Jinja, but that is only used by Sphinx and https://cove.readthedocs.io/en/update-reqs/ shows that as fine

Bjwebb (approved):

PR #1353, opened 2021-07-21
Explore:  Add new organisation referenece section.

kindly:


cove_iati/tests.py
```diff
@@ -296,10 +296,10 @@ def test_cove_iati_cli_output():
...
                     'path': '/iati-activities/iati-activity[3]/iati-identifier/text()',
                     'rule': 'identifier/text() should match the regex [^\\:\\&\\|\\?]+'},
-                   {'id': '?TZ-BRLA-5-CCC-123123-CC123',
-                    'explanation': '?TZ-BRLA-5 does not match the regex ^[^\\/\\&\\|\\?]+$',
-                    'path': '/iati-activities/iati-activity[3]/reporting-org/@ref',
-                    'rule': 'reporting-org/@ref should match the regex [^\\:\\&\\|\\?]+'},
+                   #{'id': '?TZ-BRLA-5-CCC-123123-CC123',
+                    #'explanation': '?TZ-BRLA-5 does not match the regex ^[^\\/\\&\\|\\?]+$',
+                    #'path': '/iati-activities/iati-activity[3]/reporting-org/@ref',
+                    #'rule': 'reporting-org/@ref should match the regex [^\\:\\&\\|\\?]+'},
```
Bjwebb: Should this be removed rather than commented out?

Bjwebb:
Tick mark is in the wrong place:

Screenshot from 2021-09-14 15-42-20

cove_iati/lib/iati.py
```diff
@@ -447,3 +479,93 @@ def organisation_identifier_count(tree):
     unique_identifiers = list(set(identifiers))
 
     return len(unique_identifiers)
+
+
+ACTIVITY_PREFIX = '/iati-activities/iati-activity'
```
Bjwebb, line 492: Have we thought about whether we want to check org files too?

cove_iati/lib/iati.py
```diff
@@ -447,3 +479,93 @@ def organisation_identifier_count(tree):
...
+
+
+def check_activity_org_refs(tree):
+    root = tree.getroot()
+    
+    publishers = requests.get("https://codelists.codeforiati.org/api/json/en/ReportingOrganisation.json").json()
+    registration_agency = requests.get("https://codelists.codeforiati.org/api/json/en/OrganisationRegistrationAgency.json").json()
```
Bjwebb: As discussed earlier, it would be good to deal with these JSON files being offline, in a way that doesn't break the whole results page.

cove_iati/lib/iati.py
```diff
@@ -447,3 +479,93 @@ def organisation_identifier_count(tree):
...
+                if org in reporting_org:
+                    continue
+
+                # get possible prefixes by initial substrings of ref. This is faster than checking ref
+                # against all prefixes in registry.
+                found_prefix = None
+                for possible_prefix in set(org[0:n] for n in range(4, len(org)+1)):  # no prefix shorther than 4
+                    if possible_prefix in org_prefixes:
+                        found_prefix = possible_prefix
+                        break
```
Bjwebb: Finding found_prefix happens even if it's not needed, because org is in publisher_codes

Bjwebb: But, the actual logic here looks good to me!

Bjwebb (changes requested):
I've left a few specific cmoments, but generally looking good!

cove_iati/lib/iati.py
kindly: yes but that can be at a seperate time.  It requires checking different fields.

kindly:
@Bjwebb all changes are done, with a test!

cove_iati/lib/iati.py
```diff
@@ -447,3 +487,118 @@ def organisation_identifier_count(tree):
...
+
+
+def check_activity_org_refs(tree):
+    root = tree.getroot()
+
+    publisher_request = requests.get("https://codelists.codeforiati.org/api/json/en/ReportingOrganisation.json")
+    registration_agency_request = requests.get("https://codelists.codeforiati.org/api/json/en/OrganisationRegistrationAgency.json")
+
+    if publisher_request.status_code != 200 or registration_agency_request.status_code != 200:
+        return {"error": "Unable to fetch data to do organisation checks", 'not_found_orgs_count': 0}
```
Bjwebb: This doesn't catch ConnectionError or SSLError (which I tried out by editing my hosts file).

cove_iati/lib/iati.py
```diff
@@ -447,3 +487,118 @@ def organisation_identifier_count(tree):
...
+                if org in reporting_org:
+                    continue
+
+                # get possible prefixes by initial substrings of ref. This is faster than checking ref
+                # against all prefixes in registry.
+                found_prefix = None
+                for possible_prefix in set(org[0:n] for n in range(4, len(org)+1)):  # no prefix shorter than 4
+                    if possible_prefix in org_prefixes:
+                        found_prefix = possible_prefix
+                        break
```
Bjwebb: This is a duplicate of the code below, should this instance be deleted?

Bjwebb:
Build failed because coveralls is down. The tests pass, so I'm merging this anyway.

Bjwebb (approved):

PR #1364, opened 2021-10-14
Non Embedded Codelists: Load from remote source

ghost:
https://github.com/OpenDataServices/cove/issues/1301


ghost:
Haven't fully tested for now.

This removes the 'filename' key for non embedded code lists - but as far as I could tell, it wasn't actually being used anywhere.

I wasn't sure whether to remove `@lru_cache()` from `non_embedded_codelists`. The problem with leaving it is that cache doesn't have an expiration so it would be possible for a server process to start up, run for weeks and never get new code lists (mind you, that still is better than our current situation, and uwsgi can be set to cycle workers regularly). But if we remove it every upload will be parsing the XML from the requests cache from scratch. Maybe that's a performance hit we don't care about but thought I'd mention it.

Also I wasn't sure what to do with the new request.raise_for_status() - should we catch it and mark the section as unavailable, like the other place REQUESTS_SESSION_WITH_CACHING is used?

Maybe also call the function on server thread start, so anything downloaded and cached is done straight away? If not it will happen on first user request, and we might as well try to avoid slowdown for users.

cove_iati/settings.py
```diff
@@ -125,3 +125,64 @@
...
+    'ResultType': 'https://codelists.codeforiati.org/api/xml/ResultType.xml',
+    'ResultVocabulary': 'https://codelists.codeforiati.org/api/xml/ResultVocabulary.xml',
+    'SectorCategory': 'https://codelists.codeforiati.org/api/xml/SectorCategory.xml',
+    'SectorVocabulary': 'https://codelists.codeforiati.org/api/xml/SectorVocabulary.xml',
+    'Sector': 'https://codelists.codeforiati.org/api/xml/Sector.xml',
+    'TagVocabulary': 'https://codelists.codeforiati.org/api/xml/TagVocabulary.xml',
+    'TiedStatus': 'https://codelists.codeforiati.org/api/xml/TiedStatus.xml',
+    'VerificationStatus': 'https://codelists.codeforiati.org/api/xml/VerificationStatus.xml',
+    'Version': 'https://codelists.codeforiati.org/api/xml/Version.xml',
+}
```
Bjwebb, line 188: Could be nicer to construct these urls based on the keys?

Bjwebb (approved):

cove_iati/settings.py
ghost: Yes, but my thinking was that implies we know the URL pattern will stay consistent and I thought that wasn't something I could guarantee.

ghost:
Tested by putting ` <activity-scope code="3" />` (good)  and ` <activity-scope code="333333333333333" />` (bad) into the sample activity XML file.

ghost:
Deployed to dev & tested, deployed to live & tested

PR #1369, opened 2022-05-18
Add "Transform" section, with merge indicators by title

Bjwebb:
There's now a dev deploy of this at http://merge-indicator.dev.cove.opendataservices.coop/data/f8074830-45f1-4caa-a3b6-ebb4a2c21d30

I've added a "Transform" block that's collapsed by default:

Screenshot from 2022-05-18 14-05-11

This is what it looks like expanded:

Screenshot from 2022-05-18 14-05-22

After hitting the Run button:

Screenshot from 2022-05-18 14-20-19

This is all on one line, with a view to adding extra Transforms in the future.

@stevieflow How does this look? Are you happy with this Transform block being visible to all users? If you want we can write some text to go into the block to explain what it is.

### DIFF
--- a/cove_iati/templates/cove_iati/explore.html
+++ b/cove_iati/templates/cove_iati/explore.html
@@ -419,12 +419,13 @@
     <div id="transform-body" class="collapse {% if suffix %}in{% endif %}">
       <div class="panel-body">
         <p>
-          Merge indicators by title:
+          <h4>Merge indicators by title</h4>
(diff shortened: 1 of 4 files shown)


Bjwebb:
@rhiaro This depends on this iati-utils PR https://github.com/OpenDataServices/iati-utils/pull/9

stevieflow:
I tested this and confirm it works as expected

> @stevieflow How does this look? Are you happy with this Transform block being visible to all users? If you want we can write some text to go into the block to explain what it is.

Yes, all good.  Ill write this text in a new issue, with a couple of other observations - but this does not need to block the PR

rhiaro (approved):

PR #1377, opened 2023-03-01
Dokku: Start and Python: Upgrade to 3.11 & upgrade all packages

ghost:


ghost:
I'm puzzled that `sudo apt-get install build-essential libxml2-dev libxslt1-dev python3-dev` from the readme doesn't seem to be needed?

Bjwebb:
> I'm puzzled that `sudo apt-get install build-essential libxml2-dev libxslt1-dev python3-dev` from the readme doesn't seem to be needed?

So, none of these are needed if we use the binary wheel for lxml.

For now we're building our own wheel, so libxml2-dev and libxslt-dev are required (and are installed by the current Dockerfile). I assume build-essential is installed by default. Similarly, it seems like the python install comes with the necessary development files.

Bjwebb (approved):

PR #1393, opened 2024-08-23
requirements: Update flattentool

Bjwebb:


cove_iati/tests_functional.py
simon-20, line 1: Just curious as to why this custom namespace has moved here?

simon-20 (approved):

cove_iati/tests_functional.py
Bjwebb: This is a result of handling namespaces that do/don't cover a whole activity: https://github.com/OpenDataServices/flatten-tool/commit/7e12343ff6735a6c4ddc231519304cb153d40e6c

The easiest way of implementing it was to create all parent elements after their children, and pass the namespace information to the parents. lxml then outputs the namespaces in the most "efficient" places possible. A side effect of this is that namespaces that were on the activity level are now on the root element.

simon-20: Ah, that's helpful, thanks.

Bjwebb:
I will merge/deploy this on Tuesday.

PR #1399, opened 2025-04-16
Relicense from AGPL to BSD 3-clause

jarofgreen:


jarofgreen:
There is a test issue unrelated to this - I propose temporarily removing restrictions so this can be merged and looking at tests in other PR

